PR #52, opened 2017-08-07
Time logging

dapperjapper:
Added functionality to track how long building targets takes, and to display that information with the `times()` function.

BTW: Thank you for writing drake! We're using it in my office and it has been a LIFESAVER for big data processing.


wlandau-lilly:
Fantastic idea! I would love to add this feature. Do you think people might also be interested in hashing/storage times? I am wondering about workflows with large imported objects/files.

wlandau-lilly:
Also, it is gratifying to hear that drake is working out well for you and your colleagues.

dapperjapper:
@wlandau-lilly In my usage, we don't directly reference imported objects or files very often, so it's not something we run into. Also, hashing is something that is less directly affected by the development process so it's not clear that it could be optimized by end users. For me, I'm hoping to use timing to identify data processing bottlenecks and chip away at them. But there also doesn't seem to be a downside to tracking hashing/storing times! It would give a more complete picture for sure.

wlandau-lilly:
@dapperjapper That makes sense, and I suppose we could always add more timing later. I like what you have done so far, and thank you for respecting the existing style and conventions.

Just a few minor comments:
- Do you think we should call your new function `build_times()`? Just thinking out loud here, but it might help future-proof the code and guide new users. `times()` could be an alias (`times <- build_times <- function(...){...}`).
- Along those lines, it might be nice to call the `storr` namespace `"build_times"`.
- `cache$list(namespace = "build_times")` might be safer than a `tryCatch()` block with `cache$list()`.
- I think tracking times is like tracking progress. While implementing `progress()` (previously `status()`), I learned that
  1. I needed to `cache$clear(namespace = "progress")` before every serious call to `make()` to avoid mixing up old info with new info, but
  2.  in `config()`, I needed to leave the `"progress"` namespace alone because `plot_graph()` (and `dataframes_graph()`) rely on progress info. 
- It would be great to have unit testing of (i) and (ii) for times. You might weave it into [tests/testthat/test-basic.R](https://github.com/wlandau-lilly/drake/blob/master/tests/testthat/test-basic.R) to make use of existing calls to `make()`.

wlandau-lilly:
...thought I admit that I do use `tryCatch()` a lot. It can be hard to think of a way around it.

dapperjapper:
Thanks for the feedback! I haven't worked with storr before, so I was kinda naive with how I was using it. I've renamed the function to build_times, and added unit tests. However, I'm not sure it's good to clear the times cache before every build..

Wouldn't that remove build times from targets that are already up-to-date, and that won't be rebuild by the current build? Then, they are lost until the associated target is clean()ed and rebuilt. I would much prefer that build times store the time from the most recent build of that single target, even if that specific round of building is a lil old.

wlandau-lilly:
@dapperjapper You are absolutely right about not clearing build times. And fortunately, that makes things a lot simpler and less error-prone. Although, I am on the fence as to whether `clean(target)` should remove the build times of `target`. It would be a simple matter of adding `"build_times"` to the collection of `storr` namespaces in the `for` loop of `uncache()`. What do you think?

wlandau-lilly:
Also, I do not think we need `config$cache$del(key = target, namespace = "build_times")` in `build()`. If a build quits in error, it might be nice to have some hint about how long it would have taken.

dapperjapper:
I think if you aren't uncache()ing the progress namespace when you clean(), maybe you shouldn't uncache() build times? I think they are conceptually in the same category.

wlandau-lilly:
In that case, let's leave the build times alone in `clean()`. I am just about ready to merge, but I have another couple favors to ask. 
1. If you are willing, please feel free to add yourself as a contributor in the [DESCRIPTION](https://github.com/wlandau-lilly/drake/blob/master/DESCRIPTION) (`role = "ctb"` in `person()`). In my opinion, you have definitely earned it. 
2. It might be nice to mention `build_times()` in the "Useful functions" sections of the README and vignettes next to the other cache functions. Also, maybe a `roxygen2` examples block?
```r
#' @title Function \code{build_times}
#' ...
#' @examples
#' \dontrun{
#' load_basic_example()
#' make(my_plan)
#' build_times()
#' }
```

dapperjapper:
Yay! Thanks for the support.

wlandau-lilly:
Hmm... I'm not sure what happened in the code coverage report. It seems to be reading one of your earlier commits by mistake. Maybe bump the version in the DESCRIPTION to 4.1.0 and commit again?

wlandau-lilly:
Perfect! Thanks again for the great work.

wlandau-lilly:
Related to #70:  `rehash_file()` and `store_target()` compute new hashes, so it might be worth logging these times too. I think that would cover everything we would need to time.

By the way, I'm loving `build_times()`. The next release will show them in `plot_graph()` (#57).

wlandau-lilly:
FYI: I changed the time logging to cover all of `build()`. This means we are now factoring in hashing and storage, not just the execution of the commands. I was seeing overly optimistic results from `predict_runtimes()`, and this change seemed like the cleanest solution. Plus, it will now be much easier to run performance studies that compare the practical impact of different hashing algorithms.

Issue #42, opened 2017-06-07
Additional modes of parallelism


wlandau-lilly:
[zeromq.org](http://zeromq.org/)

wlandau-lilly:
Also [Hadoop](http://hadoop.apache.org/) and [sparklyr](https://spark.rstudio.com/).

wlandau-lilly:
From @mrchypark, see [airflow](https://github.com/apache/incubator-airflow) and [luigi](https://github.com/spotify/luigi).

mrchypark:
others in [this repo](https://github.com/pditommaso/awesome-pipeline)

wlandau-lilly:
For Spark, it seems like [sparklyr::spark_apply()](https://spark.rstudio.com/) would be ideal. We could wrap it into a [worker for parallel_stage()](https://github.com/wlandau-lilly/drake/blob/master/R/parallel.R#L9) like [worker_parLapply()](https://github.com/wlandau-lilly/drake/blob/master/R/parLapply.R#L21) or [worker_mclapply()](https://github.com/wlandau-lilly/drake/blob/master/R/mclapply.R#L6).

wlandau-lilly:
I just talked to a colleague, and Spark may not be the best option here. [sparklyr::spark_apply()](https://spark.rstudio.com/) somehow might come into play for #77, but it does not need to. Spark is apparently not designed for general parallelism tasks, it is more about parallelism when you're splitting up a large data file.

On the other hand, [Docker](https://www.docker.com/) may be like `Makefile` parallelism, but more reproducible and robust. It may even help solve #76 with deployment hooks. @mrchypark, what do you think?

mrchypark:
@wlandau-lilly good point. I agree docker is good choice. it's because you can monitor and kill processes. How can I help you?

wlandau-lilly:
[...]

Issue #40, opened 2017-05-19
Improve code style


wlandau-lilly:
Should use `formatR::tidy_dir()` and `lintr::lint_package()` to conform to the [style guide](http://adv-r.had.co.nz/Style.html).

wlandau-lilly:
`tidy_source()` and `tidy_dir()` remove many complaints from `lint_package()`, but not all of them, and I think the output ends up less readable in some ways. If I were to fix the style, I would need to manually go through each source file, example, and vignette regardless of how I preprocess with `formatR`. I will do this if someone expresses a strong enough preference, but otherwise, I do not think it is worth the time.

wlandau-lilly:
I think it's worth keeping this one on the table. It would be enough to follow [Hadley's style guide](http://adv-r.had.co.nz/Style.html).

wlandau-lilly:
I would add this issue to the next CRAN milestone, but I am not sure I will work on it myself until #21 is fixed.

wlandau-lilly:
Also, `dataframes_graph()` is getting pretty big and ugly. But again, I probably will not work on it until #21 is fixed.

AlexAxthelm:
I can start on this, if you want. Spacing/formatting is pretty easy for me, and my editor runs `lintr` automatically, so this would actually help me spot errors in code, rather than lint warnings. 😄 

I know you have some changes waiting to merge (re: #21 / # 22 ), but I can start with any files that you haven't changed recently, and then once `master` is up to date, run through the rest.

wlandau-lilly:
@AlexAxthelm that would be an amazing help! Here are the files safe to modify in the `R` folder:

- clean.R
- console.R
- current.R
- debug.R
- dependencies.R
- deprecate.R
- envir.R
- find.R
- generate.R
- graph.R
- lapply.R
- Makefile.R
- make.R
- mclapply.R
- outdated.R
- plan.R
- progress.R
- read.R
- test.R
- timestamp.R
- zzz.R

and in `tests/testthat`:

- test-command-changes.R
- test-envir.R
- test-examples.R
- test-full-build.R
- test-generate.R
- test-import-file.R
- test-import-object.R
- test-intermediate-file.R
- test-Makefile.R
- test-namespaced.R
- test-other-features.R

And `inst/examples/basic/basic.R` is also fair game if you work in a mention of `build_times()`.

AlexAxthelm:
No problem. I’ll get started on these soon.

[...]

Issue #25, opened 2017-04-13
Automatically generate a workflow plan data frame from arbitrary R code.


wlandau-lilly:
Just something to dream about.

wlandau-lilly:
This just might be possible after all with `CodeDepends` (`readScript()` and `getInputs()`). See duncantl/CodeDepends#14.

wlandau-lilly:
This just might be possible after all with `CodeDepends` (`readScript()` and `getInputs()`). See @clarkfitzg's example at duncantl/CodeDepends#14.

wlandau-lilly:
In fact, I will most likely solve this with `CodeDepends` ([CRAN](https://CRAN.R-project.org/package=CodeDepends), [GitHub](https://github.com/duncantl/CodeDepends)). This feature now seems within reach.

wlandau-lilly:
I have been thinking about this one intermittently for a long time. Even with `CodeDepends`, I am not sure this feature is likely to work very well with most scripts. The scripts where it will work will end up looking a lot like `plan()`, but without the generative templating powered by `expand()` and `evaluate()`. I do not see much added value. Closing for now, will reopen if anyone sees a need and a clean implementation.

#9, however, is much easier and much more useful. I do plan to stick with that one.

wlandau:
@lawremi, thanks for bringing this up this issue yesterday. I almost about it. cc @gmbecker,  your expertise is ideal here. A naive approach would be to just let the default functionality of `CodeDepends` handle the construction of a `drake` plan from a script, but I am not well versed in the edge cases that are likely to crop up when users take it and run too far.

Related idea: construct a `drake` plan from a `knitr`/R Markdown document with an extra `knit(tangle = TRUE)` preprocessing step.

gmbecker:
@wlandau So my read on this is that CodeDepends could very easily generate a plan in the simple case. You have outputs (which can be targets) and the expressions and the inputs.

If you want people to be able to have targets that have more than one expression (I don't know how easy this is currently in drake) that can be complicated, but in the basic one epxression per target case CodeDepends can be used pretty much as is, I think.

I don't know of any situations in which CodeDepends wouldn't get the right _output_ for target identification.

clarkfitzg:
I use CodeDepends for something similar in makeParallel. Relevant code: https://github.com/clarkfitzg/makeParallel/blob/master/R/inferGraph.R

wlandau:
[...]

PR #56, opened 2017-08-09
Reproducible random numbers

AlexAxthelm:
This PR is intended to increase reproducible, by having `make(plan)` preserve `.Random.seed` when building elements.

Includes test file with tests for comparing built and rebuilt objects, and ensuring that objects with same command (generated by `expand`) yield distinct results.

Note: I am getting inconsistent results from `devtools::test()`. Sometimes it will run without any errors, others it will have a lot, even when run sequentially. When I do get errors, it's almost never the same tests as the last run. I'm hoping that it's a quirk of my system, or that I am missing a package that is called in a test, because I get the same behavior on `master` as well. I'm submitting this PR, and hoping that travis can test better than me.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -32,7 +32,8 @@ Imports:
     testthat,
     tools,
     utils,
-    visNetwork
+    visNetwork,
+    withr
 Suggests:
     abind,
     knitr,
--- a/R/make.R
+++ b/R/make.R
@@ -160,7 +160,9 @@ make = function(plan, targets = drake::possible_targets(plan),
      config$graph = delete_vertices(config$graph, v = delete_these)
      if(parallelism == "Makefile") parallelism = default_parallelism()
   }
-  get(paste0("run_", parallelism), envir = getNamespace("drake"))(config)
+  withr::with_preserve_seed(
+    get(paste0("run_", parallelism), envir = getNamespace("drake"))(config)
+    )
   if(return_config) return(config)
   else invisible()
 }
--- a/tests/testthat/test-reproducible.R
+++ b/tests/testthat/test-reproducible.R
@@ -1,4 +1,4 @@
-context("reproducible")
+context("reproducible random numbers")
 
 test_that("Objects are reproducible", {
   dclean()
@@ -8,14 +8,14 @@ test_that("Objects are reproducible", {
     y = runif(20),
     z = rnorm(20)
     )
-  make(data)
+  make(data, verbose = FALSE)
   old_x <- x
   old_y <- y
   old_z <- z
 
   # Oh no, I've accidentally deleted some data that needs to be reproducible
   clean(x, y, z)
-  make(data)
+  make(data, verbose = FALSE)
   expect_identical(x, old_x)
   expect_identical(y, old_y)
   expect_identical(z, old_z)
@@ -30,7 +30,7 @@ test_that("Objects are distinct", {
     x = runif(20)
     )
   data_exp <- expand(data, values = c("a", "b"))
-  make(data_exp)
+  make(data_exp, verbose = FALSE)
 
   expect_false(identical(x_a, x_b))
 
@@ -51,7 +51,7 @@ test_that("Sequential objects reproduce correctly", {
     methods,
     data = first_stage
     )
-  make(rbind(first_stage, second_stage))
+  make(rbind(first_stage, second_stage), verbose = FALSE)
   old_x <- x
   old_xbax_x <- xbar_x
   old_y_y <- y_x
@@ -59,7 +59,7 @@ test_that("Sequential objects reproduce correctly", {
   # How do I keep deleting my data like this? So clumsy.
   clean(x, xbar_x, y_x)
 
-  make(rbind(first_stage, second_stage))
+  make(rbind(first_stage, second_stage), verbose = FALSE)
   expect_identical(old_x, x)
   expect_identical(old_xbax_x, xbar_x)
   expect_identical(old_y_y, y_x)
--- a/tests/testthat/test-reproducible.R
+++ b/tests/testthat/test-reproducible.R
@@ -8,14 +8,26 @@ test_that("Objects are reproducible", {
     y = runif(20),
     z = rnorm(20)
     )
-  make(data, verbose = FALSE)
-  old_x <- x
+  make(
+    data,
+    envir = dbug()$envir,
+    parallelism = testopts()$parallelism,
+    jobs = testopts()$jobs,
+    verbose = FALSE
+    )
+  old_x <- envir$x
   old_y <- y
   old_z <- z
 
   # Oh no, I've accidentally deleted some data that needs to be reproducible
   clean(x, y, z)
-  make(data, verbose = FALSE)
+  make(
+    data,
+    envir = dbug()$envir,
+    parallelism = testopts()$parallelism,
+    jobs = testopts()$jobs,
+    verbose = FALSE
+    )
   expect_identical(x, old_x)
   expect_identical(y, old_y)
   expect_identical(z, old_z)
@@ -30,7 +42,13 @@ test_that("Objects are distinct", {
     x = runif(20)
     )
   data_exp <- expand(data, values = c("a", "b"))
-  make(data_exp, verbose = FALSE)
+  make(
+    data_exp,
+    envir = dbug()$envir,
+    parallelism = testopts()$parallelism,
+    jobs = testopts()$jobs,
+    verbose = FALSE
+    )
 
   expect_false(identical(x_a, x_b))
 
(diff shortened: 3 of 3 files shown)


wlandau-lilly:
This is a super elegant approach to a problem I thought would be harder to solve. 

Just so I understand: are you trying to ensure that each target starts with the same seed? I think that would be okay, I just have a mild concern that RNGs may no longer be pseudo-random if the seed is reset too often.

I think the testing problems may be caused because `withr::with_preserve_seed()` uses the global environment.

```r
with_preserve_seed
## function (code)
## {
##     old_seed <- get_valid_seed()
##     on.exit(assign(".Random.seed", old_seed, globalenv()), add = TRUE)
##     code
## }
```

On the other hand, a lot of functions like `make()` begin with

```r
make = function(..., envir = parent.frame()){
  ...
}
```

In my testing, `envir` is a `new.env(parent = globalenv())`. You may have to define your own version of `with_preserve_seed()` that uses `envir` instead of `globalenv()`.

wlandau-lilly:
Never mind about my original concern about resetting the seed. I will need to think about this more, but there is only one reset per call to `make()`, so RNGs should still be pseudo-random throughout the workflow.

AlexAxthelm:
The main idea is that in a script of vignette, we could have something along the lines of:

```r
set.seed(123)
make(plan)
```
 where plan has random number generation, or (in my case) sampling, and we would get the same results, regardless of where or when it is run.

AlexAxthelm:
I hadn't considered that `envir` might have a different `.Random.seed`. Maybe I can use `withr::with_seed()` instead, and pull the seed from there.

AlexAxthelm:
Upon further investigation of `.Random.seed`, it seems that there it only exists in `globalenv()`. Even when I attempt to explicitly set it inside an environment, all the RNG functions call on the one in `globalenv()`

```r
> ls(globalenv(), all = TRUE)
character(0)
> a <- rnorm(1)
> ls(globalenv(), all = TRUE)
[1] ".Random.seed" "a"
> foo <- .Random.seed
> rnorm(3)
[1] -0.07880674 -1.15956053 -0.09012208
> rnorm(3) # Note this is diefferent
[1] -0.7723668 -1.6987008  0.5537592
> envir <- new.env()
> envir[[".Random.seed"]] <- foo #Explicitly set the environment .Random.seed
> ls(envir, all = TRUE)
[1] ".Random.seed"
> f <- function(){
+ force(envir) # Same as we see in drake::make()
+ rnorm(3) # If calling using the envir .Random.seed, we should see same results as before
+ }
> f() #But no.
[1]  0.8525371 -0.7058269  0.1934947
> .Random.seed <- foo
> rnorm(3)
[1] -0.07880674 -1.15956053 -0.09012208
> .Random.seed <- foo
> f() # even with the force(envir), it's still using global .Random.seed
[1] -0.07880674 -1.15956053 -0.09012208
>
```

This leads me to believe that `with_preserve_seed()` is the right way to go. I'm guessing that the testing problems on my local machine are particular to my machine.

wlandau-lilly:
What are the specs of your system and R version? Maybe I can reproduce the test errors. In my experience, the CRAN check servers do not always work properly, especially R-devel on Windows, so I need to be extra diligent to make sure all the checks pass.

After reading your tests more closely, the problem you are solving is much clearer to me: even if the user does not remember to set the seed beforehand, `make()` should restore the seed for the next call to `make()` so that targets can be repaired and recovered with the same seed automatically.

Speaking of different forms of parallelism, almost all of drake's unit tests are configured to use `envir = dbug()$envir`, `parallelism = testopts()$parallelism`, and `jobs = testopts()$jobs`. Everything needs to work under all combinations of these settings, and seeds may not carry over easily, especially with `Makefile` parallelism in a non-global environment. Would you mind passing these arguments to `make()` in `test-reproducible`? 

Bonus points if you know a clean, automated way to run `testthat::test("drake")` under different scenarios instead of manually changing `R/debug.R` and `R/test.R` each time. (Maybe with [mockr](https://github.com/krlmlr/mockr) somehow?)

AlexAxthelm:
Working on implementing the arguments into `make()` commands now.

What different scenarios would you want to test?

As far as my local environments, I'm on Windows10, and I also have a setup with BashOnUbuntuOnWindows

## Win 10 (vanilla)
```R
> sessionInfo()
R version 3.4.1 (2017-06-30)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 14393)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252
[2] LC_CTYPE=English_United States.1252
[3] LC_MONETARY=English_United States.1252
[4] LC_NUMERIC=C
[5] LC_TIME=English_United States.1252

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

other attached packages:
[1] fcuk_0.1.21

loaded via a namespace (and not attached):
[1] compiler_3.4.1     magrittr_1.5       parallel_3.4.1     tibble_1.3.3
[5] Rcpp_0.12.11       rlang_0.1.1        stringdist_0.9.4.4 purrr_0.2.2
```

## BashOnWindows
```R
> sessionInfo()
R version 3.4.1 (2017-06-30)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 14.04.5 LTS

Matrix products: default
BLAS: /usr/lib/libblas/libblas.so.3.0
LAPACK: /usr/lib/lapack/liblapack.so.3.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C
 [9] LC_ADDRESS=C               LC_TELEPHONE=C
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

loaded via a namespace (and not attached):
[1] compiler_3.4.1
```

wlandau-lilly:
Thanks, @AlexAxthelm. Unfortunately I do not have access to Windows 10. Are the tests failing because the seed is not recovered, or are there other errors?

I think I digressed when I brought up testing configurations. I made a separate issue in #58.

AlexAxthelm:
I've been working on this today, and I think we should take it off of the 4.10 milestone. I've tried a couple of different variations on the `withr::with_seed()` theme, and they all only work when `jobs = 1` which is a pretty severe limitation to enforce. Looking through the[ RNG documentation](https://stat.ethz.ch/R-manual/R-devel/library/parallel/doc/parallel.pdf)(page 5) in the `parallel` package does not leave me hopeful. That doesn't even touch the `Makefile` parallelism.

I vote we close for now, and turn this into an issue with the "difficult" tag.

wlandau-lilly:
Useful information. I agree, it seems like the current approach will not work for parallel execution. This makes me think we should handle the seeds on a target-by-target basis. We would need to wrap `withr::with_seed()` around the body of `build()` and use the `storr` cache to keep track of

1. The original seed at the start of `make()`.
2. The ending seed of each target.

For the starting seed of each subsequent target, we could choose among the ending seeds of the dependencies using some deterministic rule that does not depend on execution order.

I think this is doable, but as you say, difficult. I am okay with keeping this PR around if you would like to work on it. Otherwise, we can create a separate issue for another time.

At the moment, my tentative plans for drake are these (plus any production-ready pull requests that come in the meantime)

- 4.1.0: mostly bugfixes for which I have patched issues locally, plus all the accepted PRs so far.
- 4.1.1: any straightforward, low-risk work on style and cleanup not in 4.1.0: linting, testing, #40, lingering bugs, etc.
- 4.2.0: this issue, plus #9, #25, and maybe #42.
- 4.2.1: bugfixes from 4.2.0

AlexAxthelm:
I'll probably try again on this in a week or two. My current line of thought also has me wanting to cache the seed as the first line of `make()`, but then I think I want to pass that (same value) to each of the parallel sessions.

A thought that I had was to store the .`Random.seed` at the beginning of `make()`, and then call that exact value from within each parallel session, which would mean we don't have to worry about tracking before/after conditions. Importantly, I would probably end up using a numeric hash of `.Random.seed` since the actual value of the seed is not as important as the fact that it can be set/reset/recalled (I'm ignoring the use case where somebody tries to get the same results *without* `drake` as they do with). Hashing the seed would let me salt the hash with the target name, ensuring that targets with identical commands would get different (reproducible) seeds, despite having the same "base seed". 

Not entirely sure what the best way to implement this will be. I'll take a shot a passing through `storr`, but my suspicion is that it will end up with as an element of `config`.

wlandau-lilly:
Sounds much cleaner and easier to implement. Do you think we will still keep the promise of pseudo-randomness if we reset the seed this way for each target? We probably do, my concern is mild.

I would just add the seed to either `config` or the cache's `"session"` namespace, either should be fine.

wlandau-lilly:
@AlexAxthelm what are your plans for this issue? Did you close the PR to start fresh later?

AlexAxthelm:
Deleted the branch by mistake while getting a little too happy with cleaning my git branches.

wlandau-lilly:
Completed/replaced by #109. @AlexAxthelm, sorry to close your PR. The ideas are all great, I simply implemented them.

AlexAxthelm:
Thank you so much for wrapping this up! I"m glad that I was able to help.

PR #70, opened 2017-08-18
More verbose about hashing

dapperjapper:
This pull request is probably bad code and is mostly just a proof of concept. 

On my machine, I've been seeing a big pause between the "import" phase and the "build" phase of my makes. I did a little research into what my machine was doing during this pause, and it seems to be mostly hash-checking targets. 

I think it would be good for drake to be a lil more verbose when making, because otherwise it looks like it is taking a long time to accomplish the last item on the console, which was just importing a small function. This way, there is a good way to diagnose wait times.

### DIFF
--- a/R/current.R
+++ b/R/current.R
@@ -5,7 +5,10 @@ should_build = function(target, hash_list, config){
 }
 
 target_current = function(target, hashes, config){
-  cat(paste("comparing hash for", target), "\r") 
+  # When overwriting console lines with \r, longer target names that precede
+  # shorter target names won't be overwritten. So we pad with spaces with
+  # sprintf to clear our these spaces.
+  cat("comparing hash for", sprintf("%-50.50s", target), "\r") 
   if(!(target %in% config$inventory)) return(FALSE)
   if(!file_current(target = target, hashes = hashes, config = config)) 
     return(FALSE)
--- a/R/current.R
+++ b/R/current.R
@@ -5,6 +5,8 @@ should_build = function(target, hash_list, config){
 }
 
 target_current = function(target, hashes, config){
+  #print(paste("comparing hash for", target))
+  cat(paste("comparing hash for", target), "\n") #" \r") 
   if(!(target %in% config$inventory)) return(FALSE)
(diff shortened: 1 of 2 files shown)


dapperjapper:
Alternatively, once drake is done importing something, you could have it do

```r
cat("\rimported", target, "\n)
```

to erase and update the last written console text, and provide better UX!

wlandau-lilly:
@dapperjapper I think you are on the right track here. Not only does `hash_list()` check lots of little files in `.drake/`, but it also computes the md5 sum of any new or changed external file targets/imports. (Does this sound likely to affect you?) Extra logging sounds great, and it has the added bonus of delimiting the parallelizable stages (except for proper targets in Makefile parallelism).

I will think about using `"\r"` to overwrite the current line of output, though I do find it awkward sometimes when the cursor is on the same line as the output being printed.

wlandau-lilly:
I thought about this for a little bit more just now, and I think the next release will have the following in `parallel.R`:

```r
parallel_stage <- function(graph_remaining_targets, worker, config) {
  config <- inventory(config)
  remaining_targets <- V(graph_remaining_targets) %>%
    names %>% intersect(config$targets)
  candidates <- next_targets(graph_remaining_targets)
  console_parallel_stage(candidates, remaining_targets, config)
  hash_list <- hash_list(targets = candidates, config = config)
  build_these <- Filter(candidates,
    f = function(target)
      should_build(target = target, hash_list = hash_list, config = config))
  hash_list <- hash_list[build_these]
  if (length(build_these))
    worker(targets = build_these, hash_list = hash_list,
      config = config)
  delete_vertices(graph_remaining_targets, v = candidates)
}
```

and in `console.R`:

```r
console_parallel_stage <- function(candidates, remaining_targets, config){
  if (!config$verbose) return()
  cat(color("check", "slateblue2"), " ",
    length(candidates), " item",
    ifelse(length(candidates) == 1, "", "s"),
    "\n", sep = "")
}
```

That way, we will see shorter lines, nice coloring, and the number of items (targets or imports) that are actually being checked in the current phase.

But regardless of the actual message, I am glad to have something printed at the beginning of each parallelizable stage.

PR #72, opened 2017-08-23
Linting, Part 1

AlexAxthelm:
I'm trying to keep the linting PRs pretty atomic, so the whole process will be broken into several, as I complete files.

There are a lot of commits for this PR, because I've tried to keep each commit limited in scope to a single function, in case we need to revert one, and to make `blame` easier later.

In addition to the default checks for `lintr`, I've also made some style choices that I believe improve readability. 

* I've put any oneliners following an `if` or `for` inside curly braces. This was a bit of an issue for "oneliners" that actually were across multiple lines because of `%>%`. 
* Forced a newline after `%>%` to highlight additional functions.
* I've put parens at the end of functions that are inside a pipechain, to make them identifiable at a glance as functions.  
* If a function call was too long to fit on a single line, I've put each argument on its own line. Additionally, there are some calls that I applied this to, even though they would all fit, for consistency with similar calls.
* I adjusted indentation to `vim` standard (`ggVG=`)
* On a few functions, I named the first argument, since the rest were named as well.

There might be some more stylistic choices that I've missed, feel free to let me know if you don't like any of them, and I can revert them.

I think that all I have done is formatting work, but it's possible that I dropped a comma or something somewhere, and broke the whole thing. So, I have deliberately not touched the tests in this PR, but I have a few of them finished on my local machine. I think I want to finish the functions that I can, and let Travis and Appveyor run them against tests that we know that I didn't screw up.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -7,6 +7,8 @@ Authors@R: c(
     email = "will.landau@lilly.com", role = c("aut", "cre")),
   person(family = "Clarkberg", given = "Jasper",
     email = "jasper@clarkberg.org", role = c("ctb")),
+  person(family = "Axthelm", given = "Alex",
+    email = "aaxthelm@che.IN.gov", role = c("ctb")),
   person(family = "Eli Lilly and Company", role = "cph"))
 Description: A solution for reproducible code and 
     high-performance computing.
--- a/R/clean.R
+++ b/R/clean.R
@@ -116,5 +116,6 @@ uncache <- Vectorize(function(target, path, search){
       cache$del(target, namespace = space)
     }
   invisible()
-},
-"target")
+    },
+  "target"
+  )
--- a/R/clean.R
+++ b/R/clean.R
@@ -65,17 +65,20 @@ clean_everything <- function(
   search
   ){
   empty(path, search)
-  if(destroy) {
+  if (destroy) {
     destroy(path, search)
   }
   invisible()
 }
 
-destroy = function(path, search){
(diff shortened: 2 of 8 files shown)


wlandau-lilly:
@AlexAxthelm thank you for your work on this! I do plan to merge relatively soon. I will have to manually resolve merge conflicts for my local copies of `debug.R` and `console.R`, but it's not a huge deal since I know this is just a matter of linting.

AlexAxthelm:
Cool. You should probably go with your local copies of any conflicts, and I can re-lint those files later.

wlandau-lilly:
Sounds great. I actually already linted them myself, and it was easy to add back the originals.

wlandau-lilly:
Merged. Don't worry about the code coverage reduction.

wlandau-lilly:
Confirmed: a couple extra tests raised code coverage back up to 100% for my local workspace. Thanks again for your help!

Also, I just added non-failing lints to my local copy's `.travis.yml`. From my brief experience so far, I expect many false alarms (e.g. "undefined globals" for some non-exported functions), so advice would be helpful when the time comes.

AlexAxthelm:
Awesome! When we fold in `lintr` to Travis, we can make a `.lintr` file with the line:

```r
linters: with_defaults(object_usage_linter = NULL)
```
This disables the linter which checks for exported objects. I forgot that I have this as part of my defaults in vim.

See also : [Project Configuration](https://github.com/jimhester/lintr#project-configuration)

Issue #77, opened 2017-08-28
drake friendly alternative to dplyr::do


AlexAxthelm:
Most of my development is done on my local machine, which chokes on datasets that most people wouldn't even think of as "big data". One of the reasons that I was drawn to `drake` in the first place was the idea that I could segment my data, and let my machine work on each part individually, making checkpoints along the way so that if it does choke, I don't have to re-run the whole thing.

So I wrote a series of functions that give me similar functionality to `dplyr::do()`, but in drake. Specifically, I've written `drake_split()` and `drake_unsplit()` commands, which output plans. This allows me to ~~horribly misuse~~ *creatively exploit* the `analyses()` function to process my small data chunks in parallel, including tasks such as validating (using `assertr`) and cleaning. 

I've got the bones of this "issue" solved, but I don't want to turn it into a PR until I can write some tests and maybe a vignette or example (hopefully later this week). Mostly, I'm opening this to solicit requests for functionality and to put it on the roadmap for later milestones.

### Current/Planned features:
- [ ] Splits automatically:
  - [x] `data.frame`
  - [x] `tibble`
  - [ ] `data.table` (It might do this, I haven't tested at all)
- [x] Recombines automatically
- [x] Plays well with `magrittr` pipe
  - [ ] Plays well with multi-step pipechains (Ex: `mtcars %>% head %>% drake_split()`)
- [x] Respects dplyr groups
  - [x] Attempts to distribute groups evenly (by row count) among slices
- [x] Can apply multiple steps of analysis to each parallel slice (I use this to validate, clean, and then analyze each slice)
- [ ] Offer a choice of using the `dplyr` of `base` functions for splitting/binding
- [ ] Documentation
  - [ ] `Roxygen` docs
  - [ ] Example script
  - [ ] Vignette
- [ ] Unit Tests for each feature

wlandau-lilly:
@AlexAxthelm this sounds really clever! Would you mind sharing a little demo? I am very curious, and I am not 100% sure how your new feature set plays out.

wlandau-lilly:
@AlexAxthelm So this is a way to break up a big data frame and process the chunks into multiple targets? That sounds super cool, I love it!

More generally, I wonder if it's possible to leverage `dplyr` itself somehow. That could be a separate feature set, if it can be done.

One small suggestion: it may be slightly more future-proof to use `evaluate()` directly rather than `analyses()`, though `analyses()` does get the point across more easily.

AlexAxthelm:
[...]

PR #79, opened 2017-08-29
Initial work on splitting

AlexAxthelm:
This is a start on building the `do` functionality that I mentioned in #77. 

I expect this to fail on Travis and Appveyor, since I used `test_with_dir()` rather than the vanilla `test_that()`. Also there is no documentation, but the last test shows an (impractical, but fast running) example case, for sequential work. The graph for the plan is shown here:
image

I have no idea why the parallel pipelines are graphed with a curve to them, but notice that they all stem from the (imported blue) `mtcars` on the left, and reconnect at `mtc_cleaned` near the right. I chose to run the recombined dataframe through an `lm`, so that I woulnd't have to worry about `test_that` complaining about differently ordered rows, even though they should be the same.

### DIFF
--- a/R/split.R
+++ b/R/split.R
@@ -75,6 +75,40 @@ split_list <- function(
   return(split_list)
 }
 
+analyses_split <- function(
+  split_plan,
+  methods,
+  wildcard = "..datasets..",
+  # Magic wildcards will expand the wildcard automatically, such that the
+  # wildcard for the second step does not need to be prefixed with the
+  # targetname from the first
+  magic_wildcards = TRUE
+  ){
+  valid_targets <- split_plan$target[!grepl(
+    pattern = "^split_list(.*, splits = .*)$",
+    x = split_plan$command
+    )]
+  if (magic_wildcards){
+    methods_targets <- methods$target
+    wild_prefix <- paste0(dplyr::lag(methods_targets), "_")
+    # remove the initial wildcard prefix, since we want the first argument to
+    # find the original splits, not NA_splits
+    wild_prefix[1] <- ""
+    # Actually substutute the wild_prefix into for wildcard
+    methods <- evaluate(
+      plan = methods,
+      wildcard = wildcard,
+      values = paste0(wild_prefix, wildcard),
+      expand = FALSE
+      )
+  }
+  evaluate(
+    plan = methods,
+    wildcard = wildcard,
+    values = valid_targets
+    )
+}
+
 drake_unsplit <- function(
   plan,
   out_target,
--- a/R/split.R
+++ b/R/split.R
@@ -1,0 +1,131 @@
+drake_split <- function(
+  .data,
+  out_target,
+  # this is here so you can use the expected rows, if data does not yet exist
+  data_rows = nrow(.data),
+  splits = ceiling(nrow(.data) / 1e6)
+  ){
+  name <- deparse(substitute(.data))
+  # Check if .data is an object in a magrittr pipeline.
+  if (name == "."){
+    name <- find_lhs_name(.data)
+  }
+  # Ensure that splits is an integer:
+  splits <- ceiling(splits)
+  digits <- ceiling(log10(splits))
+  # deparse target names (pre-application)
+  slice_targets <- paste(
+    "slice",
+    name,
+    stringr::str_pad(seq(splits), pad = "0", side = "left", width = digits),
+    sep = "_")
+  # determine the number of rows in each split (also an integer)
+  split_rows <- ceiling(data_rows / splits)
+  if (dplyr::is_grouped_df(.data)){
+    grouplist_target <- paste("slice", name, "grouplist", sep = "_")
+    slice_targets <- c(
+      grouplist_target,
+      slice_targets
+      )
+    slice_commands <- c(
+      paste0("group_list(", name, ", splits = ", splits, ")"),
+      paste0(
+        name,
+        "[dplyr::group_indices(", name, ") %in% ",
+        grouplist_target, "[[", seq(splits), "]], ]"
+        )
+      )
+  } else {
+    split_seq_low <- seq(
+      from = 1L,
+      by = split_rows,
+      length.out = splits
+      )
+    split_seq_high <- seq(
+      from = split_rows,
+      by = split_rows,
+      length.out = splits
+      )
+    slice_commands <- paste0(
+      "dplyr::slice(",
+      name, ", ",
+      split_seq_low, ":", split_seq_high,
+      ")")
+  }
+  slice_plan <- data.frame(target = slice_targets, command = slice_commands)
+  return(slice_plan)
+}
+
+group_list <- function(
+  .data,
+  splits
+  ){
+  # determine the size of each group
+  data_groups <- dplyr::group_size(.data)
+  names(data_groups) <- seq_along(data_groups)
+  # ensure split_rows is an integer
+  split_rows <- ceiling(sum(data_groups) / splits)
+  group_list <- vector("list", length = splits)
+  for (i in data_groups){
+    push_to <- which.min(lapply(group_list, sum))
+    push_value <- data_groups[which.max(data_groups)]
+    group_list[[push_to]] <- c(group_list[[push_to]], push_value)
+    data_groups[which.max(data_groups)] <- NA
+  }
+  # Extract just the names, since those are the actual values we want.
+  group_list <- lapply(group_list, names)
+  group_list <- lapply(group_list, as.numeric)
+  return(group_list)
+}
+
+drake_unsplit <- function(
+  plan,
+  out_target,
+  last_method = gsub(
+    x = tail(plan$target, 1),
+    pattern = "^(.*)_slice.*$", "\\1"
+    ),
+  plan_type = "analysis" #c("analysis", "split")
+  ){
+  stopifnot(is.character(last_method))
+  stopifnot(plan_type %in% c("analysis", "split"))
+  if (missing(out_target)){
+    out_target <- deparse(substitute(plan))
+  }
+  stopifnot(is.character(out_target))
+  if (plan_type == "analysis"){
+    bind_targets <- paste0(
+      grep(
+        pattern = last_method,
+        x = plan$target,
+        value = TRUE
+        ),
+      collapse = ", "
+      )
+  } else {
+    bind_targets <- paste0(
+      last_method, "_",
+      split_plan[["target"]],
+      collapse = ", "
+      )
+  }
+  unsplit_plan <- data.frame(
+    target = out_target,
+    command = paste0("rbind(", bind_targets, ")")
+    )
+  return(unsplit_plan)
+}
+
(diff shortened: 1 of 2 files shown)


wlandau-lilly:
I really like what you have done so far. Tidyverse or not, it is super easy to split up a data frame. Some preliminary thoughts:

### Big datasets as targets

What if the big dataset is a target later in the workflow? There is a little hiccup in the interface if it is not yet in the environment.

```r
drake_split(x, data_rows = 4, splits = 2)
## Error in inherits(x, "grouped_df") : object 'x' not found
x <- 1
drake_split(x, data_rows = 4, splits = 2)
##      target              command
## 1 slice_x_1 dplyr::slice(x, 1:2)
## 2 slice_x_2 dplyr::slice(x, 3:4)
```

And it may be expedient to specify the dataset with a quoted target name.

```r
drake_split("targ", data_rows = 4, splits = 2)
##           target                   command
## 1 slice_"targ"_1 dplyr::slice("targ", 1:2)
## 2 slice_"targ"_2 dplyr::slice("targ", 3:4)
```

`readd()` solves this problem with a `character_only` argument.

```r
readd <- function(target, character_only = FALSE, ...){
   ...
   if(!character_only) target = as.character(substitute(target))
   ...
}
```

In this case, it may also be unwise to hard-code the row indices passed to `dplyr::slice()`, since the actual number of rows in the target may be unpredictable.

### Suggested number of splits

The default `splits = ceiling(nrow(.data) / 1e6)` seems like a great choice based on my experience with R's performance, but I wonder if we are missing a more sophisticated metric or community standard. And to my knowledge, nobody has rigorously explored the performance tradeoff between target size and the number of targets for drake specifically. This point will not stop me from accepting the PR, just thinking out loud.

AlexAxthelm:
I agree, This is mostly something I have been using semi-interactively, so I've been using it on targets that already exist. The pipe operator fails outright if the object doesn't exist (`kwyjibo %>% head()`).

I agree that `character_only` is the way to go. for passing object that don't exist yet.

I probably will end up making an additional target, like the `grouplist` target that I've used for the grouped frames, and I'll probably end up using `base` selection, rather than dplyr then: `frame[rowlist[[X]], ]`

wlandau-lilly:
I forgot to address your point about the vertical positions of nodes. I struggled with this for 4.0.0, but could not get `visNetwork` to respond to all of my instructions in the `x`, `y`, and `level` columns of the `nodes` data frame. I hope `visNetwork` solves this, I vaguely remember posting an issue. The `layout` argument of `drake::plot_graph()` and `drake::render_graph()` offers a tiny bit of flexibility in the meantime, but it is not a complete solution.

AlexAxthelm:
Did some work on this branch today. 

Major Updates:
* Changed the structure of `drake_split()` plans, so they have the same structure, regardless of if the `.data` is grouped or not. 
  * The actual subsetting/slicing is taken care of when `make` is run, rather than when the plan is made, meaning that the plan can be made before the data exists. (Must still manually define the number of splits in that case, so an estimated number of rows is still helpful to have.)
* Created the terribly named `analyses_split`, which is a variant of `analyses()` that makes writing the parallel plans a bit easier. 
  * seriously, the function name should be changed. 
  * `magic_wildcards` argument name should probably also be changed. But it is a binary flag which takes the `methods` plan, and auto-magically treats the commands as ordered, and will build the full plan with the splits inserted, and dependencies upon the previous line in `methods`
    * weird to explain, but the example-test in the updated `test-split.R` has the flag enabled, allowing `methods` to just have `..datasets..`, none of the weird `step_one_..datasets..` that I needed before
* moved tests back to `test_that` so that travis can actually try them.
* Using base selection, rather than `dplyr::filter`
* Changing phrasing from "slice" to "split" (Partially done)

AlexAxthelm:
Upon seeing the CI results, I'm noticing that I inadvertently put a dependency on `dplyr` in here. I hadn't actually noticed that it wasn't in there before. It's kind of a hard requirement for checking on dplyr groups, which seem like a natural breakpoint for data frames.

My suggestion would be either:
1. add `dplyr` to `suggests`, and not check for groups if it's not installed, since the user couldn't have them anyway.
2: remove `magrittr` from `Imports`, replacing it with `dplyr`, and import the pipe operator from there.

I don't have much of a preference, especially since dplyr is a super common package to have installed already, and if someone has dplyr, then they also have `magrittr`.

wlandau:
I am not sure about (1) versus (2) just yet. (If (2), we should make sure to update the unfortunate `load_packages_parLapply()`. Otherwise, `parLapply` functionality will fail on Windows R-devel.)

The reason I am not sure is that you opened such a huge door with the idea of splitting. We could split evenly across the rows, use `dplyr::group_by()`, leverage columns as split variables some other way, split things that are not data frames, etc. In fact, it would be really cool to formalize this API to allow different splitting methods to be added in modular fashion. Here, you might have separate functions for uniform row splitting vs `group_by()` row splitting, and they could share common standards in naming, style, behavior, etc.

I also wonder: if `group_by()` could dictate a split during `make()`, we might think about how to manage the number of splits differently. Here are some choices.

1. Distribute the data over a fixed pre-specified number of splits. Here data chunks would not correspond 1:1 with the targets. Some targets might get multiple data chunks, some would get nothing.
1. Amend the workflow plan during a call to `make()`. We can only really know the best number of splits after we actually have a big dataset that needs splitting. What if we wait until then, modify the graph, and then proceed with a new workflow? This behavior could also help data analyses where a major strategic decision / pivot is made programatically somewhere in the middle (choose your own adventure).

(2) sounds extremely difficult, and it may not even be possible. It may even require a successor to `drake` just as `drake` arose from some shortcomings of [remake](https://github.com/richfitz/remake).

wlandau-lilly:
I am not sure about (1) versus (2) just yet. (If (2), we should make sure to update the unfortunate `load_packages_parLapply()`. Otherwise, `parLapply` functionality will fail on Windows R-devel.)

The reason I am not sure is that you opened such a huge door with the idea of splitting. We could split evenly across the rows, use `dplyr::group_by()`, leverage columns as split variables some other way, split things that are not data frames, etc. In fact, it would be really cool to formalize this API to allow different splitting methods to be added in modular fashion. Here, you might have separate functions for uniform row splitting vs `group_by()` row splitting, and they could share common standards in naming, style, behavior, etc.

I also wonder: if `group_by()` could dictate a split during `make()`, we might think about how to manage the number of splits differently. Here are some choices.

1. Distribute the data over a fixed pre-specified number of splits. Here data chunks would not correspond 1:1 with the targets. Some targets might get multiple data chunks, some would get nothing.
1. Amend the workflow plan during a call to `make()`. We can only really know the best number of splits after we actually have a big dataset that needs splitting. What if we wait until then, modify the graph, and then proceed with a new workflow? This behavior could also help data analyses where a major strategic decision / pivot is made programatically somewhere in the middle (choose your own adventure).

(2) sounds extremely difficult, and it may not even be possible. It may even require a successor to `drake` just as `drake` arose from some shortcomings of [remake](https://github.com/richfitz/remake).

AlexAxthelm:
Splitting seems to me to be a pretty key functionality, even if it's not baked right into `drake`. I don't know if this is a feature that should be a separate extension package (I doubt that is the way to go though).

The latest push has support for splitting evely either across rows, or, if groups exist, it will split on those instead, and attempt to fill the splits as evenly as possible (one of the reasons that I picked the `starwars` dataset is because for hair_color there are a few groups that are very large, and many groups with 1-3 entries).

## Extending `drake_split()`
I agree that we ~could~ should have an extensible function that checks the type of the object being split (I haven't even thought about lists). I'll make some changes on my end to accommodate this. Currently, I'm treating the grouped and ungrouped dataframes, tibbles, etc. as mostly being the same though, and the `split_list` function outputs a list of row indices that I select upon. I suppose the same method could be used for splitting a list. 

In extending the `split_<thing>`, it seems prudent to start putting a `...` into the `drake_split` arguments, so we can include some non-standard ways of defining split criteria.

## Auto-splitting
As for splitting being something that happens natively in `make()`, I am opposed to this. `drake` is friendly, but definitely a package for power-users. I think that splitting automatically would remove both transparency and plan customization from the end-user. 

For starters, we would have to pick a default number of splits (number of `jobs`? number of `jobs` that aren't being used on this step?  `nrow()/1e6`? something else?) Whatever default we pick for `make()` would end up being a bad choice for a lot of people, either because it leave each split too large for a user's machine (my issue), or because it would over-parallelize, and introduce too much overhead. 

Additionally, I think that splitting needs to be explicitly requested by the user. This is especially true in the case of `grouped_df`s. Ungrouped stuff won't change order, but because I'm trying to split the work evenly, I end up moving some groups around, so that when all of the splits are reunited, they are almost certainly not in the same order. `starwars` by hair_color, will yield groups `10, 7, 4` first (since they are the largest), with everything else coming after them. This not not an unexpected behavior for anybody working with groups (I believe that `dplyr::do` also takes liberties with group ordering), but I certainly don't want it to be an automatic thing. I know that I work with `lead()` and `lag()` frequently enough that the idea of anything changing my row ordering without me know makes me a bit twitchy.

wlandau-lilly:
Good points about auto-splitting. You have convinced me that we should choose the number of splits in advance, particularly to maintain predictability and control for end users. And I am a little less worried about `drake`'s ability to keep up with the times.

Also, I just saw `drake_unsplit()`, and I forgot to mention that `gather()` does the same thing. In fact, `summaries()` optionally calls `gather()`, with an option to turn it off.

wlandau-lilly:
Also, I had assumed we would build these splitting methods into drake itself. The way our discussion is going makes me lean toward replacing `magrittr` with `dplyr` in `Imports:`, etc.

AlexAxthelm:
Thanks for pointing out `gather` I had forgotten that it exists.

`drake_unsplit()` does in fact have the same function as `gather(gather = "rbind")`

R/split.R
```diff
@@ -0,0 +1,162 @@
...
+      )
+  }
+  evaluate(
+    plan = methods,
+    wildcard = wildcard,
+    values = valid_targets
+    )
+}
+
+drake_unsplit <- function(
```
wlandau-lilly: As I previously mentioned, I think we could rely on `gather_plan()` here. Maybe that means removing `drake_unsplit()` entirely, I do not know.

R/split.R
```diff
@@ -0,0 +1,162 @@
...
+    wild_prefix[1] <- ""
+    # Actually substutute the wild_prefix into for wildcard
+    methods <- evaluate(
+      plan = methods,
+      wildcard = wildcard,
+      values = paste0(wild_prefix, wildcard),
+      expand = FALSE
+      )
+  }
+  evaluate(
```
wlandau-lilly: For #147, many functions have been deprecated and renamed due to potential name conflicts. For example, `evaluate()` is now `evaluate_plan()`.

wlandau:
@AlexAxthelm, since you gave permission, I am closing this PR in favor of #233. I'm glad GitHub refers back to these things because your ideas are great reference material.

Issue #64, opened 2017-08-15
Predict the runtime of the next call to make().


wlandau-lilly:
Sketch:

```r
predict_runtime <- function(plan, parallelism, jobs, ...){
  # What to we need to build?
  run_these <- outdated(plan = plan, ...)

  # How long did those targets take before?
  times <- build_times()[run_these,] # Need to deal with brand new targets somehow.

  # What kind of parallelism do we have?
  nodes <- dataframes_graph(plan = plan, parallelism = parallelism, jobs = jobs, ...)
  
  # Account for targets not yet timed.
  # Account for parallelism.
  # Return estimated total runtime.
}
```

@dapperjapper, back to our conversation from #52, this could a reason to record all the times, not just build times. I am not sure.

dapperjapper:
Looks cool! I can't work on it over the next 2 weeks, but it's a good idea.

wlandau-lilly:
Fixed. See [predict_runtime()](https://github.com/wlandau-lilly/drake/blob/0444b8f4a3813330e0c4cbb63e648e810c03b0aa/R/predict_runtime.R#L63), [rate_limiting_times()](https://github.com/wlandau-lilly/drake/blob/0444b8f4a3813330e0c4cbb63e648e810c03b0aa/R/predict_runtime.R#L187), and the new [timing vignette](https://github.com/wlandau-lilly/drake/blob/master/vignettes/timing.Rmd).

PR #80, opened 2017-08-29
Predicting build time

dapperjapper:
I'm not sure if this is what you had in mind, but here's my initial implementation (issue #64)! I'm open to critique and revisions on this. Here's what the output looks like for the debug plan:

```
> predict_runtime(plan, envir = envir, from_scratch = T)
import i
import a
import b
import c
import saveRDS
import 'input.rds'
import readRDS
import j
import h
import g
import f
Build stage 1 
  Targets: myinput 
  Est build time: 0s 
Build stage 2 
  Targets: nextone 
  Est build time: 0s 
Build stage 3 
  Targets: yourinput 
  Est build time: 0s 
Build stage 4 
  Targets: combined 
  Est build time: 0s 
Build stage 5 
  Targets: 'intermediatefile.rds' 
  Est build time: 0s 
Build stage 6 
  Targets: final 
  Est build time: 0s 

TOTAL BUILD TIME: 0s 
  0 untimed targets (never built)
  (assuming max_useful_jobs)
  (not including hashing and storage time [yet])
```


wlandau-lilly:
@dapperjapper Looks like a great start. I like that you print an estimate separately for each parallelizable stage. This feature will take some iteration, and I have a couple preliminary suggestions. 

- We may want to return more than just a single time estimate. Maybe a list or data frame of the stage-specific information would help. 
- What about a range of times rather than a single time estimate? Sorry this did not occur to me before, but it may not be right to always assume either `max_useful_jobs()` or serial execution.
- You could use `drake:::multiline_message()` to keep console output clean if there are many targets.
- We may want to include a verbosity flag (for the imports).
- And of course, docs and tests.

Also, I should mention that I have done a lot of development on my own this month, and none of it has been approved for release from my company yet. I will merge your feature into the upstream copy when it is ready, and I will integrate the changes into my local development copy, but the merge conflicts for the latter will be a headache. Your contribution will keep its functionality, but your code will probably look different in the end.

wlandau-lilly:
@dapperjapper I just pushed 4.1.0 (on its way to CRAN). Sorry, but there are some new merge conflicts.

wlandau-lilly:
@dapperjapper I really like your idea to include `untimed_method`. For my own projects, I plan to supply `min()` and `max()` to predict best-case and worst-case scenarios.

That got me thinking: what if we allowed users to manually pre-specify their own runtime predictions for each target? Actual stored build times would override them. I have a couple different interface alternatives in mind:

- Allow an optional third column, something like `time` or `runtime_prediction`, for the workflow `plan`.
- Allow `untimed_method` to be a named list, named vector, or data frame of target-level runtime predictions.

What do you think?

R/timing.R
```diff
@@ -67,3 +67,124 @@ build_times <- function(
...
+                            untimed_method = mean,
+                            build_times = data.frame(target = character(0),
+                                                     elapsed = duration(numeric(0)),
+                                               stringsAsFactors = F),
+                            digits = 0,
+                            verbose = FALSE,
+                            config = NULL,
+                            ...){
+  
+  if (missing(config))
```
wlandau-lilly: For `missing()`, see a [tweet from @gaborcsardi](https://twitter.com/GaborCsardi/status/894625773994336256) and the ensuing thread.

wlandau-lilly:
I really appreciate what you have done here, and I have thought a lot about it. It has helped me think about what a good prediction would look like and what the user might want. I just sketched out my own first attempt in the [`predict_runtime` branch](https://github.com/wlandau-lilly/drake/tree/predict_runtime). I plan to debug it, add unit tests, and write a whole vignette on timing.

wlandau-lilly:
Like I said, you really helped me think about #64. But I went ahead and wrote my own code, so I will not be accepting the exact code in this PR.

PR #90, opened 2017-09-21
Verbose about loading dependency targets

dapperjapper:
This is another bottleneck for my usage of drake, so it would be nice to have output here that shows why drake is stalling. I also generalized your console_parallel_stage function for this purpose! This slightly changes the output of the previous console_parallel_stage "check n items" output, to print some target names.

### DIFF
--- a/R/console.R
+++ b/R/console.R
@@ -18,7 +18,7 @@ console_verb_targets <- function(targets, verb, config, color = "slateblue2"){
     return(invisible())
   }
   cat(color(verb, color), " ", n, " item",
-    ifelse(n == 1, "", "s"), ": ", head(targets, 1), 
+    ifelse(n == 1, "", "s"), ": ", head(targets, 1) %>% crop_text(length = 50), 
     ifelse(n == 1, "", ", ..."), "\n", sep = "")
 }
 
--- a/R/console.R
+++ b/R/console.R
@@ -18,8 +18,9 @@ console_verb_targets <- function(targets, verb, config, color = "slateblue2"){
     return(invisible())
   }
   cat(color(verb, color), " ", n, " item",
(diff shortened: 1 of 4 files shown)


dapperjapper:
Example output:

```
> make(plan, "lil_target")
check 2 items: head, internal_library::load_data
import head
import internal_library::load_data
check 1 item: medium_target
check 1 item: lil_target
load 1 item: medium_target
target lil_target
```

wlandau-lilly:
@dapperjapper Thank you for the idea. I will look at it closely as soon as I have the time. Right now, I am in the middle of a whirlwind of other updates, but I do think it is important to print console messages in key points where drake appears to stall.

wlandau-lilly:
By the way: in de54b3a61241776714383a501df06493135dc389, I add [mclapply-driven parallelism](https://github.com/wlandau-lilly/drake/blob/master/R/parallel.R#L26) to `hash_list()` and `loadd()`. So this bottleneck, and hashing for that matter, should be slightly less annoying on non-Windows platforms.

wlandau-lilly:
Spot on! I might change some names, but I really like this.

wlandau-lilly:
You gave me another idea: as of a8d6b5b00b2b2db021577795f311c9a46336e3e4, `drake` [also logs unloading](https://github.com/wlandau-lilly/drake/blob/a8d6b5b00b2b2db021577795f311c9a46336e3e4/R/envir.R#L35). (At every parallelizable stage, `drake` [unloads](https://github.com/wlandau-lilly/drake/blob/a8d6b5b00b2b2db021577795f311c9a46336e3e4/R/envir.R#L40) [targets that are no longer needed in the present or future](https://github.com/wlandau-lilly/drake/blob/a8d6b5b00b2b2db021577795f311c9a46336e3e4/R/envir.R#L31).) The process is quick, but I still want to draw attention to it because memory efficiency is a major selling point of this package.

PR #104, opened 2017-10-13
Fix #103

### LINKED ISSUE #103: Revert #6
After taking a break and thinking it over, I anticipate serious problems with #6. It makes projects overreactive and brittle, shortening shelf lives and triggering unexpected builds when downloading an old project to another system (similar to richfitz/remake#176, which is otherwise not a problem for drake). I think it would be better to **not** track packages as dependencies, regardless of how a package is fingerprinted.

Packrat is a much more appropriate solution to package reproducibility,  and it **lengthens** project shelf lives. Docker should take care of reproducibility across systems. We should recommend those tools, drake does not excel in that space.

We will need to update the images in the READ ME and vignettes, which will make this repo bigger, but it can't be helped.

wlandau:


wlandau:
Still need to fix the [RawGit](http://rawgit.com/) links after merging.

PR #108, opened 2017-10-14
Clean up internals

wlandau:
- Add more and better default arguments to user-side functions.
- Clarify the role of the `config` argument in functions that take one (`max_useful_jobs()`, `outdated()`, `plot_graph()`, ...)
- Clean up `dataframes_graph.R` and `dataframes_graph_utils.R` by relying more on the `config` argument. Objects are loaded into `config` to avoid overlarge argument lists in internal functions.


wlandau:
Also puts @dapperjapper's notes on devtools in the caution vignette (#95).

PR #160, opened 2017-11-20
Switch jobs to config$jobs in outdated

kendonB:
These are incorrectly set to the default 1 in prepare_distributed


wlandau-lilly:
Thanks, @kendonB!

PR #213, opened 2018-02-01
add double quotes to illegal target characters

kendonB:
I ran into a situation where evaluate_plan added these in the middle of the string


wlandau:
Good call, double quotes should not be in target names. I will merge when the checks pass.

wlandau:
I just pushed a commit to fix this. A test needed to be changed too.

Issue #218, opened 2018-02-01
Pseudo-randomness is not totally reproducible


aedobbyn:
@krlmlr did all the work for me here: http://rpubs.com/krlmlr/drake-random-seed

krlmlr:
To expand a bit: It appears that the random seed isn't fixed automatically. The simple example runs a plan with random output, then reruns parts of it. The output looks different for the two cases.

I think we should be fixing the random seed for each target, for reproducibility. Issue in remake: https://github.com/richfitz/remake/issues/106.

wlandau:
@aedobbyn the basic example takes random bootstraps from mtcars. Are you looking for more randomness than that?

@krlmlr the way drake approaches reproducible random seeds is in R/random.R and tests/testthat/test-random.R. drake preserves the random seed using withr::with_seed() and salts that seed for individual targets.

Sorry I did not provide links. The internet connection crashed here and I am using my phone.

wlandau:
Also, the seed is stored in the config object, which you can recover with read_drake_config (). So drake does try to be reproducible under randomness. What I worry about is that the seed is salted deterministically for different targets based on the target name (in build_target()). Should the salting b

krlmlr:
Thanks for the pointers! The example appears to do something different than the tests: I'm not clearing everything, rather I'm deleting an intermediate target.

wlandau:
Hmm... It should still work then. I wonder what the problem is.

wlandau:
See also #56 and #109.

wlandau:
@@krlmlr target4 stays the same if I replace loadd() with readd(). So weird.

wlandau:
Kirill, in that example, did you have anything leftover in your cache from a previous make ()? I am starting to think the problem is that drake does not try to load the root seed if one is already cached. See where get_valid_seed() is called in drake_config().

wlandau:
We should also probably add a `drake_read_seed()` function. I am resistant to `drake_set_seed()` that caches a new seed, because then some targets could be built with one root seed and others could be built with another. For reproducibility's sake, I think users should have to destroy the cache to reset the seed.

wlandau:
And a new test in `test-random.R` should use `set_seed()` to reproduce the error first.

krlmlr:
In that example I cleaned the `random2` target, so I guess the cache was partially populated.

Would it work if the seed was only a function of the target name, and would be set unconditionally? Users shouldn't be worried about the random seed at all, things should just work.

[...]

PR #221, opened 2018-02-03
Add a random seed argument to make() and drake_config().

wlandau:
The random seed of a `drake` project should not depend on the user's R session. This PR makes sure the seed is reproducible and stable.

### DIFF
--- a/R/config.R
+++ b/R/config.R
@@ -41,6 +41,7 @@ drake_config <- function(
   plan = drake_plan(),
   targets = drake::possible_targets(plan),
   envir = parent.frame(),
+  seed = NULL,
   verbose = 1,
   hook = default_hook,
   cache = drake::get_cache(verbose = verbose, force = force),
@@ -96,7 +97,7 @@ drake_config <- function(
     overwrite_hash_algos = FALSE,
     jobs = jobs
   )
-  seed <- get_previous_seed(cache = cache)
+  seed <- choose_seed(supplied = seed, cache = cache)
   trigger <- match.arg(arg = trigger, choices = triggers())
   if (is.null(graph)){
     graph <- build_drake_graph(plan = plan, targets = targets,
--- a/R/make.R
+++ b/R/make.R
@@ -386,6 +386,7 @@ make <- function(
       plan = plan,
       targets = targets,
       envir = envir,
+      seed = seed,
       verbose = verbose,
       hook = hook,
       parallelism = parallelism,
--- a/R/make.R
+++ b/R/make.R
@@ -57,7 +57,7 @@
 #'   Thus, the `seed` argument must either be `NULL` or the same
 #'   seed from the project's cache (usually the `.drake/` folder).
 #'   To reset the random number generator seed for a project,
-#'   use [clean(destroy = TRUE)].
+#'   use `clean(destroy = TRUE)`.
 #'
 #' @param verbose logical or numeric, control printing to the console.
 #'   \describe{
--- a/R/make.R
+++ b/R/make.R
@@ -43,6 +43,22 @@
 #'   from `envir` and the global environment and
 #'   then reproducibly tracked as dependencies.
 #'
+#' @param seed integer, the root pseudo-random seed to use for your project.
+#'   To ensure reproducibility across different R sessions,
+#'   `set.seed()` and `.Random.seed` are ignored and have no affect on
+#'   `drake` workflows. Conversely, `make()` does not change `.Random.seed`,
+#'   even when pseudo-random numbers are generated.
+#'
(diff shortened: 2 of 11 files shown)


R/make.R
```diff
@@ -43,6 +43,22 @@
 #'   from `envir` and the global environment and
 #'   then reproducibly tracked as dependencies.
 #'
+#' @param seed integer, the root pseudo-random seed to use for your project.
```
krlmlr: Do you want to use `@inheritParams`?

R/config.R
```diff
@@ -41,6 +41,7 @@ drake_config <- function(
   plan = drake_plan(),
   targets = drake::possible_targets(plan),
   envir = parent.frame(),
+  seed = NULL,
```
krlmlr: Are you sure you want to add the new argument here?

R/make.R
```diff
@@ -324,6 +340,7 @@ make <- function(
   plan = drake_plan(),
   targets = drake::possible_targets(plan),
   envir = parent.frame(),
+  seed = NULL,
```
krlmlr: Same.

R/random.R
```diff
@@ -1,23 +1,27 @@
...
+  seed_conflict <-
+    !is.null(previous) &&
+    !is.null(supplied) &&
+    !identical(previous, supplied)
+  if (seed_conflict){
+    stop(
+      "You supplied a seed of ", supplied,
+      "to either make() or drake_config(). ",
+      "Your project already has a different seed: ", previous, ". ",
+      "Use read_drake_seed() to see for yourself.",
```
krlmlr: How do I un-bork the drake?

R/make.R
wlandau: I just learned about @inheritParams from @hadley's workshop yesterday, and I used it to get the `drake_config()` params from `make()`. For `make()` itself, where would  we inherit the params from?

R/make.R
wlandau: I get the sense that I'm not taking full advantage of `@inheritParams`. It would be *super* nice to cut down on those massive argument lists some more.

R/random.R
wlandau: Do you mean that we throw separate error messages for each of `make()` and `drake_config()`?

krlmlr: No, but we might want ti suggest how to force the new seed if that's what the user wants.

R/config.R
krlmlr: Adding `seed` in this place changes the API, if you want to be super-strict. The semantics of the call `make(a, b, c, d)` with unnamed arguments suddenly change.

One way to enforce argument naming is to place arguments after `...`, don't know if this is a good idea to do now, after the fact. Or we simply add arguments to the end, but this may distort the order.

How about an API like `drake_options() %>% set_seed(...) %>% make()`? Not right now, but eventually.

R/make.R
krlmlr: I think arguments should be documented by the lowest-level callee that exposes them.

R/config.R
wlandau: > Adding seed in this place changes the API, if you want to be super-strict. The semantics of the call make(a, b, c, d) with unnamed arguments suddenly change.

Just moved `seed` to the end in 7d69a199009ff010de4a1fd46a8f115fe728a2a7. I am willing to consider arguments after `...` or alternatively just waiting until the next major release to change the API.

> How about an API like drake_options() %>% set_seed(...) %>% make()? Not right now, but eventually.

We're almost there with `drake_config()` and `make_with_config()`. Seems like the biggest change would be to adopt `...` in place of the massive argument list in `drake_config()`. The main advantage I see is that we as the developers would be able to add or remove arguments more easily.

After all the discussion from #218, I am concerned about the confusion it would cause to go back to relying on `set_seed()`.

R/make.R
wlandau: So the inheritance should be the other way around? `make()` should inherit from `drake_config()`? Just say the word and I'll make the change.

R/random.R
wlandau: Done: c284385f13782ca3d39b00cfe70ea44c0d792733

R/make.R
wlandau: On the other hand, if we make `drake_config()` act like the `drake_options()` you suggested, then `drake_config()` will only have a `...` argument and we won't be able to inherit the params from it.

krlmlr: It looks like `drake_config()` would be the right place to document the `seed` (and other) arguments.

wlandau: I am starting to see what you mean about inheriting parameters from lower-level functions. I think I do a slightly better job in dfd34ae7dfc94411db1f8d4096ab72784a5888ed.

PR #223, opened 2018-02-03
Use roxygen2 in rstudio

krlmlr:
and make sure EOF is consistent.


wlandau:
Thanks, Kirill! I always just run `roxygen` and testing through a shortcut in the Linux terminal, and I was just using the default `drake.Rproj` without thinking about it.

PR #224, opened 2018-02-03
Add support for active bindings in loadd()

wlandau:
I have been trying to use `bindr::populate_env()` to load targets with `loadd(lazy = "binding")`, but the feature is not currently working. @krlmlr, I will need your help.


R/read.R
```diff
@@ -207,26 +212,51 @@ loadd <- function(
...
+  }
+}
+
 load_target <- function(target, cache, namespace, envir, verbose, lazy){
-  if (lazy){
-    lazy_load_target(
+  lazy <- parse_lazy_arg(lazy)
+  switch(
+    lazy,
+    eager = eager_load_target(
```
krlmlr: This feels like an S3 method `load_target()` with three implementations would be useful.

wlandau: Yeah, internally we could just assign a `lazy` class to the target name and let S3 decide among `load_target.eager()`, etc. Seems more concise. Would really shorten `store_target()` as well.

I resisted OOP in the early days because I found it hard to wrap my head around the problem `drake` tries to solve, but I am open to rethinking some data structures. The `drake_config()` output has been a list until now, but there may be advantages to making it a class. Also, I think workers and queues in the scheduler very naturally become classes.

wlandau:
I'm not sure if we need a way to remove active bindings. Is there anything special going on that interferes with the usual `rm()`?

krlmlr:
`populate_env()` works best with vectors of targets, but it also can be used with a scalar:

```r
  bindr::populate_env(
    env = envir,
    names = target,
    fun = drake::eager_load_target,
    .envir = envir,
    # the target argument will be substituted by populate_env(), we have names = target above
    cache = cache,
    namespace = namespace,
    envir = envir,
    verbose = verbose
  )
```

krlmlr:
`eager_load_target()` will need to support symbols in addition to strings for the `target` argument.

wlandau:
Thanks, Kirill! I think your hints got me unstuck. This works for me:

```r
load_basic_example()
make(my_plan, verbose = FALSE)
rm(small)
cache <- get_cache()
quick_read <- function(key, cache){
  cache$get(key = as.character(key))
}
bindr::populate_env(
  env = globalenv(),
  names = "small",
  fun = quick_read,
  cache = cache
)
head(small)

##       x    y
## 1 3.730 17.3
## 2 5.250 10.4
## 3 3.730 17.3
## 4 5.345 14.7
## 5 3.190 24.4
## 6 3.440 17.8

clean(destroy = TRUE, verbose = FALSE)
make(my_plan, verbose = FALSE)
small

##   x y
## 1 1 4
## 2 2 5
```

`eager_load_target()` and `readd()` were overkill anyway.

wlandau:
Thanks again for your help, @krlmlr!

PR #234, opened 2018-02-04
Add .drake directory to .gitignore 

uribo:
The content of this PR is simple.
In the project using drake, it is added to `.gitignore` when `drake::make()` is executed for the first time.

Using the drake environment, files and data to the `.drake` directory stored, but this is not normally included in version control by git (it is also in [this package](https://github.com/ropensci/drake/blob/master/.gitignore#L16)). 

So, in order to prevent git management of `.drake` folder by the user's unintentionally, like adding behaviors of adding various hidden files in [httr’s](https://github.com/r-lib/httr) `.httr-oauth` file and [usethis](https://github.com/r-lib/usethis) package, add to the `.gitignore`.

nzafu5qxle


R/sanitize.R
```diff
@@ -76,3 +76,35 @@ repair_target_names <- function(x){
...
+#' @title Exclude version control of the drake project cache files.
+#' @description Add drake cache directory (hidden folder named `.drake`)
+#' to `.gitignore`, when if there is not registered.
+#' @param path path to `.gitignore`. Typically below the root directory.
+#' @param verbose If TRUE, show status message.
+#' @examples
+#' \dontrun{
+#' add_ignore(path = '.gitignore', verbose = FALSE)
+#' }
+add_ignore <- function(path = NULL, verbose = FALSE) {
```
krlmlr: Could you just forward to `usethis::use_git_ignore()`? If we want to avoid a strong dependency on *usethis*, we can check if the package is available first.

R/config.R
```diff
@@ -349,6 +349,7 @@ drake_config <- function(
     packages = packages,
     prework = prework
   )
+  add_ignore(path = ".gitignore", verbose = FALSE)
```
krlmlr: @wlandau: Is `drake_config()` meant to be a pure function?

R/sanitize.R
wlandau: Yes, I would rather offload to use this as well.

R/config.R
wlandau: No, drake_config () is not pure. It initializes the cache.

wlandau:
Thanks, @uribo! I agree that this is the right thing to do b

R/config.R
krlmlr: That's not what I'd expect from a function with that name, but we can think about it later.

R/sanitize.R
uribo: @krlmlr thank you for your improve comment.

I did not know `usethis::use_git_ignore()`.
I will revert this commit for `add_ignore ()`, and replacement to `use_git_ignore()`.

@wlandau  May I add usethis (> = 1.2.0) to DESCRIPTION?

uribo:
Ummm…  Switching the part [`add_ignore()`](https://github.com/ropensci/drake/pull/234/files/099aded32d23683f8eab5fc9bd4d6d810a51cb10#diff-ee4f2069f84a82501ec5a79dd09e66bbR352) that I added to `usethis::use_git_ignore()` caused the following error in test-Makefile.R:

> Error: Current working directory, '/tmp/RtmppRSrls/drake.Rcheck',  does not appear to be inside a project or package.

When using `use_git_ignore()`, it may be difficult to clear this test.

<details>

```r
Status: 2 ERRORs, 1 WARNING, 1 NOTE
See
  ‘/tmp/RtmppRSrls/drake.Rcheck/00check.log’
for details.

R CMD check results
2 errors | 0 warnings | 1 note 
checking examples ... ERROR
Running examples in ‘drake-Ex.R’ failed
The error most likely occurred in:

> base::assign(".ptime", proc.time(), pos = "CheckExEnv")
> ### Name: drake_plan
> ### Title: Create a workflow plan data frame for the 'plan' argument of
> ###   'make()'.
> ### Aliases: drake_plan
> 
... 11 lines ...
+     local_object <- 1 + 1
+     2 + sqrt(local_object)
+   }
+ )
> small_plan
        target                                                     command
1 small_target {\n    local_object <- 1 + 1\n    2 + sqrt(local_object)\n}
> cache <- storr::storr_environment() # Avoid writing files in examples.
> make(small_plan, cache = cache) # Most users don't need the cache argument.
Error: Current working directory, '/tmp/RtmppRSrls/drake.Rcheck',  does not appear to be inside a project or package.
Execution halted

checking tests ... ERROR
  Running ‘testthat.R’
Running the tests in ‘tests/testthat.R’ failed.
Last 13 lines of output:
  ══ testthat results  ═══════════════════════════════════════════════════════════════════════════════
  OK: 420 SKIPPED: 0 FAILED: 92
  1. Error: recipe commands (@test-Makefile.R#9) 
  2. Error: no Makefile for make_imports() (@test-Makefile.R#25) 
  3. Error: prepend arg works (@test-Makefile.R#36) 
  4. Error: files inside directories can be timestamped (@test-Makefile.R#53) 
  5. Error: basic Makefile stuff works (@test-Makefile.R#73) 
  6. Error: Makefile stuff in globalenv() (@test-Makefile.R#125) 
  7. Error: packages are loaded in prework (@test-Makefile.R#160) 
  8. Error: basic example works (@test-basic.R#9) 
  9. Error: dependency profile (@test-cache.R#4) 
  1. ...
  
  Error: testthat unit tests failed
  Execution halted

checking package dependencies ... NOTE
Package suggested but not available for checking: ‘future.batchtools’
```

</details>

R/sanitize.R
wlandau: `usethis` 1.2.0 is relatively recent. This makes me think we should add it to `Suggests:` and then not add to `.gitignore` if it is not installed (and show an informative message).

wlandau:
@uribo That's really strange. `.gitignore` is not part of the `.drake` cache or anything else that `drake` currently sees, so I do not know what could be the problem.

krlmlr:
*usethis* requires the current working directory to be set to a project, can be done with `withr::with_dir()`. Does this help?

uribo:
The cause of the problem turned out. `usethis::use_git_ignore()` will stop if it is executed in a directory other than RStudio project.

https://github.com/r-lib/usethis/blob/c3e5d1447c23133768e506b0d358ffa8cbf72916/R/git.R#L65

https://github.com/r-lib/usethis/blob/da1a8c2586646e8da75be530251aad0931df6298/R/proj.R#L61-L77

If usethis is not installed, or if not using rstduio project, i will try improvements that do not add ignore.

uribo:
@krlmlr `withr::with_dir()`!!  That's a very good idea!

krlmlr:
You might need a file named `.here`, though. In *rprojroot* there's `is_remake_project`, we should add an `is_drake_project` which *usethis* would have to use and then we wouldn't need the `.here` file. But that's a rather long way to go.

uribo:
`drake_config()` is used around the package (*testthat*, *r-example*), it is difficult to apply `withr::with_dir()` to these 😢

wlandau:
I wrap the examples and tests in calls to `drake:::test_with_dir()` to try to contain side effects. This function is essentially a call to `with_dir(tempfile())` wrapped around a call to `test_that()`. This kind of technique is becoming more popular for packages that need to write files, and I am quite fond of it. It makes sure tests can be interrupted without my having to clean anything up manually, and it means I can do different testing runs on several places at once simultaneously without having to worry about conflicts. And if `testthat` ever supports parallel computing, `drake` will be ready.

Unfortunately, all this means nested calls to `withr::with_dir()` probably don't work.

uribo:
@wlandau Thank you for informative information. I will try it now.

uribo:
The `drake:::test_with_dir()` is useful. However, since [the temporary folder](https://github.com/ropensci/drake/blob/master/R/test.R#L60-L71) provided by this function is not a project directory, it seems that need to add a line to execute `usethis::create_project` in order to execute `usethis::use_git_ignore()`.

```r
test_with_dir <- function(desc, ...){
  new <- tempfile()
  dir_empty(new)
  with_dir(
    new = new,
    code = {
      set_test_backend()
      invisible(capture_output((usethis::create_project(new, rstudio = FALSE, open = FALSE))))
      test_that(desc = desc, ...)
    }
  )
  invisible()
}
```

@wlandau @krlmlr 
It may be necessary to change some of the tests, but is there any other problem with adding this behavior?

wlandau:
Hmm... not sure how I feel about this. Is an RStudio project really required? If so, maybe the best we can do is add to `.gitignore` if one already exists. A test for this could get around `withr::with_dir()` if needed, maybe with something like this.

```r
test_that("Adding to .gitignore", {
  wd <- getwd()
  on.exit(setwd(wd))
  setwd(tempfile())
  usethis::create_project()
  # Now run the test.
})
```

But to be honest, I am not sure what the right thing to do is here.

krlmlr:
@wlandau: The last code you posted is pretty much what `withr::with_dir()` is doing.

wlandau:
Yeah, I see that now. I wrote that code chunk hastily. Just trying to think of a way to get around nested calls to `withr::with_dir()`:

```r
withr::with_dir(tempfile(), withr::with_dir(tempfile(), 2))
## Error in setwd(dir = new) : cannot change working directory
```

wlandau:
I thought nested calls to `withr::with_dir()` might be the problem, but that shouldn't matter if the path to the second directory can be resolved while in the first one.

```r
dir.create("~/a")
dir.create("~/b")
withr::with_dir("~/a", withr::with_dir("~/b", 2))

## [1] 2
```

krlmlr:
A `withr::with_new_tempdir()` function might be useful, would you mind filing an issue with *withr*?

krlmlr:
The problem is that `with_dir()` needs an existing directory.

wlandau:
I would also like `withr::with_new_tempdir()` to be added to `withr`. I just appended a comment to the existing r-lib/withr#63.

wlandau:
Closing via 9e93590fe9ebc4fa321f32949475a588dc79eebb. @uribo this is a fantastic feature for large projects. I just think it is much easier to solve with a separate `.gitignore` file inside the `.drake` cache itself. `new_cache()` takes care of this now.

Issue #239, opened 2018-02-05
Consider hyper-convenient wrapper to build a single target from the console?


kendonB:
I know I've asked for something similar before, and the current solution is:

```
my_config <- read_drake_config()
my_meta <- read_drake_meta()
loadd(my_target, deps = TRUE)
drake_build(target = "my_target", config = my_config, meta = my_meta)
```
This seems like it could get wrapped up into a single user-friendly function:

```
just_build_it_really_easily_with_one_line <- function(target, config = read_drake_config(), meta = read_drake_meta()){
  loadd(my_target, deps = TRUE)
  drake_build(target = target, config, meta)
}
```

wlandau:
If you're running into this a lot, I suppose we should have an easy one-liner, maybe the following + documentation.

```r
make_single <- function(target){
  my_config <- read_drake_config()
  loadd(my_target, deps = TRUE)
  drake_build(target = "my_target", config = my_config) # `drake_build()` already gets the latest metadata from `config`
}
```

With the stream of incoming issues, I am relying more and more on pull requests from others, especially for small issues like this one.. If you have the time, I would really appreciate it.

kendonB:
I'm not sure about the function name. Does ``drake_build`` as it is need to be exported? Can this wrapper be ``drake_build`` and what was ``drake_build`` have a name change?

wlandau:
Good point. The internals are in flux right now, so I can try and free up the name.

wlandau:
Fixed via #242. It was harder than I expected. I was playing around with `drake_build()`, and it kept reverting things I was doing in the console. I had to add a `replace` argument to make sure things in the user's environment are not replaced by `loadd(deps = TRUE)`.

Issue #237, opened 2018-02-05
Add `future` support for the Makefile backend.


wlandau:
As discussed in #227, this would help us ramp up to a custom scheduler and provide some lasting advantages.

wlandau:
I no longer think this issue is necessary. I only wanted it as a way to ramp up to #227, and I think I'm [really close now](https://github.com/ropensci/drake/blob/ebd4ee149f9cddbfe73a78d2b22713afc8285d4c/R/future.R#L1) (branch [`i227-attempt2`](https://github.com/ropensci/drake/tree/i227-attempt2)). Assuming our queue is simple, so are the parallel algorithm and the data structures. I just needed to wrap my head around it.

Issue #236, opened 2018-02-05
Do not write to the cache anywhere in build_target()


wlandau:
`drake_build()` should just return a target's value, and it should not cache errors or progress on its own. This is especially important for jobs deployed to remote locations that have no access to the cache. Transferring data should be the responsibility of the `future` package.

kendonB:
I recall at one point the value was returned using future (as well as written to disk) and the process was horribly slow and memory hungry for the host. For SLURM, this (at least) doubles the I/O writing; once to bring the data back to the host process and once to write it to disk from the host process. This will be a lot of work for the host process for a large project which usually has limited resources.

> I find when running my project using future_lapply, after the slurm jobs complete, the host R process's memory usage blows up (slowly) in htop. Even if this is isn't real memory usage, it's still problematic as I'm running the host process on the shared build node. Is this the behavior you expect? Is the host process bringing back all the data to the host before writing to disk?


Ref: https://github.com/ropensci/drake/issues/115

wlandau:
So then we should think about a user-side option to cache inside or outside the future. Either way, the internals need to be cleaned up to account for all this.

kendonB:
I wonder if there's a future environment variable that indicates whether the future has access to the main disk? That might be able to shield the user from yet another argument to make

wlandau:
There may be, but I think there's also more to it than that: certain `storr` drivers are not threadsafe, and connections are lost in remote jobs. For example, `storr_dbi()` only works locally with one job. I would prefer a special setting here because keeping a master list of safe caches seems brittle.

wlandau:
I think we're good here: in #237 and #227, all we need to do is pay attention to where we use [`build_and_store()`](https://github.com/ropensci/drake/blob/ad3412b9f191b590f50613f36baf75a56e5c37bd/R/build.R#L70) versus [`just_build()`](https://github.com/ropensci/drake/blob/ad3412b9f191b590f50613f36baf75a56e5c37bd/R/build.R#L92). It should not be hard now.

Issue #233, opened 2018-02-04
DSL based on dplyr-like verbs?


krlmlr:
- Native support for verbs like `tibble()`, `crossing()`, `mutate()`, `group_by()`, `summarize()`
    - Goal: Get rid of `plan_analyses()` and `plan_summaries()`

## Sketch for "basic" example

We might want to use our own verbs, this is the current tidyverse nomenclature to communicate semantics.

```r
drake_plan(
  small = simulate(48),
  large = simulate(64),
  datasets = tibble(dataset = list(small, large)),
  regressions = tibble(reg = list(reg1, reg2)),
  analyses = crossing(datasets, regressions) %>%
    rowwise() %>% 
    mutate(result = list(reg(dataset))),
  summary_funs = tibble(fun = list(coefficients, residuals)),
  summaries = crossing(analyses, summary_funs) %>%
    rowwise() %>% 
    mutate(summary = fun(dataset, result)),
  winners = summaries %>%
    group_by(dataset, fun) %>%
    summarize(winner = min(summary))
)
```

krlmlr:
@AlexAxthelm: Just to clarify: We use these verbs only to describe the workflow, the actual execution plan will be a full expansion. But if we know that e.g. all `analyses` have been built, we don't need to look at the full expansion at all. We still need to figure out how to handle all this internally.

AlexAxthelm:
You may not have a clear answer for this, but as a design question, would each element (row) inside one of these tibble targets show separately on `outdated`, or the graph? Or would all of `analyses` need rebuilt if a single one fell outdated? Not trying to be difficult, just having trouble imagining how this will play out in execution.

krlmlr:
These questions are great!

We have an internal graph, there each target has its separate up-to-dateness. We never materialize the full internal graph, but to build `analyses` we of course look at all internal dependencies. We don't need to update those rows in `analyses` that are up to date, and we don't materialize the entire `analyses` data frame either unless the user asks for it. For visualization, we can add a "partially outdated" state if only some rows in `analyses` are stale.

AlexAxthelm:
That sounds awesome. I'm fully in support of this over #77 then. A feature that I would find useful as a user, would be to have an accessible estimate on *how* outdated a target is, e.g 7 of 9 elements need rebuilt. Other than that, I think all my concerns are met.

wlandau:
[...]

PR #241, opened 2018-02-06
Pre-work leading up to difficult parallelism issues

wlandau:
This PR fine-tunes the building and caching infrastructure, and it should make #227, #233, #237, and #239 much easier to solve. I also think it solves #236 to the full extent possible before direct implementation begins for #237. Details:

- Restructure `build_target()` so that it does not need access to the cache. Add a unit test to track.
- Do some polishing on `drake_build()` and `store_target()`. The logic is clearer, more readable, and more modular. It will help us to write a version of `drake_build()` that lets the master process do the caching (#227, #237).
- Put all the staged parallelism functions in `staged_parallelism.R`. Similarly put other functions in the files where they belong.

This PR polishes some super sensitive internals, so regular unit testing is not good enough. I will do [the pre-CRAN long-running tests on all backends](https://github.com/ropensci/drake/blob/master/tests/scenarios/all.R) and then test on an old work project before I merge this tomorrow.


wlandau:
The longer, more thorough checks look good too. Merging.

Issue #227, opened 2018-02-04
Manual scheduling


krlmlr:
- use future for all parallelism
    - get rid of `parLapply()` and `mclapply()`
    - parallelism argument is kept, forwards to the corresponding *future* implementation with a warning
- for now, only one job per future, we provide persistence later by running multiple targets in the same future
- communicating changes to the storr:
    1. job writes directly to the storr
    2. job returns result through the future, master stores result into the storr
- monitoring progress:
    - watch the storr **and**
    - check `resolved(f)`
- failing workers
    - `resolved(f)` is `TRUE`, `value(f)` throws an error
- scheduling with a priority queue (https://github.com/DataWookie/liqueueR/issues/2)
    - keys in the pq: number of unresolved upstream targets, need "decrease key" operation
    - as soon as a job finishes, "decrease key" is called for all of its downstream targets
    - a job can be run if key in pq is zero
    - job priorities help resolve ties if pq contains multiple jobs with key 0
        - one option is to use the position in the data frame as a proxy for priority if we don't have explicit priorities

## Nomenclature

- target: the object to compute
- command/job: the R code that computes a target
- future: the vehicle that computes one or more commands/jobs on a worker
- worker: the compute unit that computes futures

wlandau:
Also: let's use `lightly_parallelize()` as the engine behind `make_imports()`. For imports, the order of processing does not matter.

krlmlr:
If the keys are integers (like in our case), "insert" and "decrease key" can be implemented in O(1).

wlandau:
My main focus right now is this issue. After some sketching (which I will push to a branch as soon as I finish traveling) I think the infrastructure should emerge as an independent “future” backend separate from “future_lapply”. The new scheduler will turn out great in the end, but until we test it over several months, I do not think it should replace any of the existing backends. I think this will be okay because there will not be very much volume of code to add, and I will make sure it is merge-safe.

wlandau-lilly:
Key data structures I am planning:

- R6_worker: a persistent R6 object that can iterate over multiple transitory futures and targets, etc.
- R6_queue: a trivial queue fot targets that we will later turn into a priority queue.

wlandau:
[...]

PR #242, opened 2018-02-06
Make `drake_build()` more convenient to use.

wlandau:
Users like `drake_build()` to build individual targets interactively. This PR makes it easier. Other changes:

- Split up the build functions into `build_and_store()` and `just_build()` in preparation for #227, #236, and #237.
- Add a `replace` argument to `drake_build()` and `loadd()` so that objects already in the user's environment need not be replaced. Defaults: `FALSE` in `drake_build()` and `TRUE` in `loadd()`. Yes, the defaults do not agree, but I think this is the right thing to do.
- Update the docs and news.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # Version 5.1.0
 
+- Move `drake_build()` to be an exclusively user-side function.
+- Add a `replace` argument to `loadd()` so that objects already in the user's environment need not be replaced.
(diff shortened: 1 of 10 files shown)


wlandau:
Forgot to mention: this solves #239.

wlandau:
HPCs and old projects worked too when I checked. Merging.

Issue #208, opened 2018-01-31
Make it easier to load workflows structured as packages.


wlandau:
From #30, environments populated with `pkgload::load_all()` look like package namespaces, and that can cause problems with `drake` workflows structured as packages. From `pkgload::load_code()`, it looks `devtools:::source_many()` would do most of the work. Thanks for the tip, @hadley!

@dapperjapper  For your workflows structured as packages, is there anything else you want to load? We could try to add on more components of `pkgload::load_all()`.

We should also extend [this part of the caution vignette](https://github.com/ropensci/drake/blob/master/vignettes/caution.Rmd#take-special-precautions-if-your-drake-project-is-a-package).

dapperjapper:
I don't really understand the proposed solution, but I can clarify exactly how we're using drake within the package structure.

`R/` contains the scripts that create and process data, as well as a script `R/drake_plan.R` which contains a function that returns a drake plan object.

While we are working on the scripts in `R/`, we repeatedly hit `devtools::load_all()` to iterate and test the newest version of the code on our temporary testing data in the global environment.

When we are ready to run our code for real and save our results, we run `make(drake_plan())` after (clunkily) managing the environment weirdness created by having our functions in fake package environment.

The final product of this research project is an Rmd, so we also include a call to `make()` in a chunk in the beginning of the Rmd (while dealing with some working directory shenanigans). This is run silently every time the Rmd is knit, ensuring that it always has the latest data.

dapperjapper:
Looking at the content of `pkgload::load_all`, we don't currently use package data so that can be left out for now. However, it would be nice if drake could be cognizant of `NAMESPACE` and use that to choose libraries to load.

wlandau:
If I understand correctly, you want to 

1. load all the scripts in `R/`.
2. Make imported functionality available from any packages mentioned in `NAMESPACE`.

Is there anything else? We might be able to combine `devtools:::source_many()` with either `loadNamespace()` or `attachNamespace()` to accomplish this. I could see a function like `load_pkg_workflow()` becoming part of `drake`.

wlandau:
Thinking about this some more, I think a better solution would be to concentrate on what happens after the package is loaded. Suppose we're using `eply` as the package that implements our workflow.

```r
library(eply) # could be devtools::load_all()
```

At this point, the usual behavior is the following. 

```r
my_plan <- drake_plan(a = eply(1234))
config <- drake_config(my_plan)
vis_drake_graph(config)
```

[...]

PR #243, opened 2018-02-07
Extract imports from packages

wlandau:
This PR fully implements and tests the `extract_imports()` function mentioned in #208. I ironed out some kinks, and I am fairly confident that it now works as expected. Before merging, I still need to:

- [x] Write the `roxygen2` help docs.
- [x] Update the `caution.Rmd` vignette.
- [x] Add a new section in the `best-practices.Rmd` vignette.


wlandau:
I no longer think I need to write a whole separate example for this. The power users intense enough to package up their workflow will know what I mean from the help file of `expose_imports()` (previously `extract_imports()`).

Issue #232, opened 2018-02-04
Use language to mark input and output files


krlmlr:
```r
drake_plan(
  knit(file_input("report.Rmd"), output = file_output("report.md"), ...)
)
```

Allows storing commands as list of language objects in version +2, we don't need to distinguish between `'` and `"` anymore.

wlandau:
I will be glad when we have this feature.

What should we do about target names in workflow plans? Should we automatically use the names inside`file_output()`s and ignore what the user puts in the `target` column? If so, how do we make sure the `file_output()`s do not conflict with symbolic targets? To minimize refactoring, we could [sanitize the workflow plan](https://github.com/ropensci/drake/blob/03749ae286d77521e5c88066f42c6987d32206ef/R/sanitize.R#L1) such that the strings inside `file_output()` become double-quoted target names in the `target` column. Other approaches would require more refactoring but could be better in the long run.

AlexAxthelm:
suggestion: use a dummy target wherever there is `file_output()`, and have an explicit dependency on that dummy target for the file. Don't know how to run that without breaking old plans though.

wlandau:
It's a good thought, but I think it will interfere with the [manual scheduler](https://github.com/ropensci/drake/blob/098bc005d551eefb2ce78b1016cff188e28c9c7e/R/future.R#L1) for #227. To know whether to build a target, the scheduler checks if there are unmet dependencies, so there would be a catch 22 for a dummy target that depends on the actual file written.

wlandau:
An alternative:

```r
drake_plan(
  whatever = knit(file_input(report.Rmd), output = file_output(report.md), ...)
)

## # A tibble: 1 x 2
##   target                 command 
##   <chr>                  <chr>  
## 1 file_output(report.md) "knit(file_input(report.Rmd), output = file_output(rep …
``` 

Other thoughts:
- For file targets in `drake_plan()`, a left-hand side should be optional. (`drake_plan(x = y)` or just `drake_plan(y)`).
- In `file_input()` and `file_output()`, quoting should be optional.

wlandau:
@AlexAxthelm thinking more about your idea, it comes really close to working. Here's how I picture it. The user sets up the workflow plan without worrying about files in the `"target"` column. The symbolic LHS targets are always return values from commands, and we can reproducibly track them along with any files generated.

```
drake_plan(
  some_return_value = {
    knit(file_input("report.Rmd"), output = file_output("report.md"), ...)
[...]

Issue #205, opened 2018-01-31
get file targets working in evaluate_plan


kendonB:
the output for the target is currently:  ``'target_name.pdf'_replacement_value``

I would expect ``'target_name_replacement_value.pdf''``

hth

wlandau:
Good question. Closing because I think I addressed this in #106. Let's follow up on that thread if you have ideas for a better interface.

Incidentally, I am starting an exciting collaboration with @krlmlr, and the interface will change a lot in the future. Eventually, the [recommended workaround](https://ropensci.github.io/drake/articles/caution.html#beware-unparsable-symbols-in-your-workflow-plan-) may no longer be needed.

kendonB:
Oh superb! I hadn't seen that workaround. I look forward to testing out any
interface changes as well.

PR #258, opened 2018-02-17
Handle files with file_input(), file_output(), and knitr_input() instead of double quotes.

wlandau:
This PR is a proposed solution to what I believe to be `drake`'s highest priority issue: #232. I am eager and excited to merge with master, and I think I can say the same for others such as @noamross and @tiernanmartin. However, this is a large change, so please speak up if you concerns.

Until now, `drake` has been using single quotes to distinguish between files and non-files. This interface has major flaws:

1. It's not intuitive or self-explanatory.
2. It creates problems with workflow plan construction / wildcard templating (e.g. #205).
3. Single-quotes are removed in `parse()`, which is trouble because of how `drake` uses static code analysis.

Main changes:

- Single quotes are deprecated, but the old API still works for existing projects. Files are represented with double-quotes internally.
- Use `file_input()` in your commands and imported functions to specify a file dependency. `file_input()` and its counterparts simply accept strings and return those same strings, but `drake` treats them as flags when it comes to code analysis.
- For file targets, use `file_output()` in your commands to declare the name of the target.
- Use `knitr_input()` in your commands to name `knitr`/`rmarkdown` targets. They are like file inputs, except that `drake` analyzes the active code chunks for target dependencies (targets explicitly loaded with `loadd()` or `readd()`).
- `drake` ignores `file_output()`s mentioned in imported functions.
- `drake` treats `knitr_input()`s in imported functions as `file_input()`s.
- `sanitize_plan()` auto-fills blank target names and forces every `file_output()` to overwrite the target name.


R/check.R
```diff
@@ -96,30 +96,14 @@ warn_bad_symbols <- function(x) {
...
-    "and double-quoted strings are just ordinary strings.")
-  for (target in seq_len(length(x))) {
-    message("\ntarget: ", names(x)[target])
-    message("strings in command:\n",
-      multiline_message(drake::drake_quotes(x[[target]],
-      single = FALSE)), sep = "")
+  if (any(grepl("'", plan$target)) || any(grepl("'", plan$command))){
+    warning(
+      "Single-quoted strings found in workflow plan. ",
+      "Single-quotes used to be the way to specify file targets/imports, ",
```
krlmlr: Will all future users see this warning if they use single quotes in their command? Can we perhaps warn only if a naked string is used as a file name (because the file exists after the run)?

R/dependencies.R
```diff
@@ -90,7 +90,7 @@ deps <- function(x){
 #' con <- make(my_plan) # Run the project, build the targets.
 #' # Get some example dependency profiles of targets.
 #' dependency_profile("small", config = con)
-#' dependency_profile("'report.md'", config = con)
+#' dependency_profile("\"report.md\"", config = con)
```
krlmlr: Is this consistent with the `deps()` call in line 54 above?

R/dependencies.R
```diff
@@ -243,22 +249,180 @@ unwrap_function <- function(funct){
...
+analyze_loadd <- function(expr){
+  args <- as.list(expr)[-1]
+  targets <- unnamed_in_list(args)
+  list <- get_specific_arg(args = args, name = "list")
+  c(targets, list)
+}
+
+analyze_readd <- function(expr){
+  args <- as.list(expr)[-1]
+  targets <- unnamed_in_list(args)
```
krlmlr: Does `readd()` now have an `args` argument?

R/dependencies.R
```diff
@@ -243,22 +249,180 @@ unwrap_function <- function(funct){
...
+  target <- get_specific_arg(args = args, name = "target")
+  c(targets, target)
+}
+
+get_specific_arg <- function(args, name){
+  tryCatch(
+    eval(args[[name]]),
```
krlmlr: Do we need to evaluate this in a specific environment?

R/read.R
```diff
@@ -33,7 +33,7 @@
 #' readd(small) # Return targets 'small' from the cache.
 #' readd("large", character_only = TRUE) # Return 'large' from the cache.
 #' # For external files, only the fingerprint/hash is stored.
-#' readd("'report.md'")
+#' readd("\"report.md\"")
```
krlmlr: Consistency?

R/utils.R
```diff
@@ -26,3 +27,25 @@ braces <- function(x) {
     x
   }
 }
+
+merge_lists <- function(x, y){
```
krlmlr: Would `utils::modifyList()` or an equivalent *purrr* function help here?

R/workplan.R
```diff
@@ -149,3 +152,133 @@ drake_plan_override <- function(target, field, config){
...
+#' vis_drake_graph(config)
+#' # See also `knitr_input()`. `knitr_input()` is like `file_input()`
+#' # except that it analyzes active code chunks in your `knitr`
+#' # source file and detects non-file dependencies.
+#' # That way, updates to the right dependencies trigger rebuilds
+#' # in your report.
+#' })
+#' }
+file_input <- function(...){
+  as.character(unlist(list(...)))
```
krlmlr: `as.character(c(...))` ?

I wonder if the implementation should add an attribute or a class, we could also add the `"fs_path"` class for nicer printing. (Need to import *fs*, though.)

R/workplan.R
```diff
@@ -149,3 +152,133 @@ drake_plan_override <- function(target, field, config){
...
+#' vis_drake_graph(config)
+#' # See also `knitr_input()`. `knitr_input()` is like `file_input()`
+#' # except that it analyzes active code chunks in your `knitr`
+#' # source file and detects non-file dependencies.
+#' # That way, updates to the right dependencies trigger rebuilds
+#' # in your report.
+#' })
+#' }
+file_output <- function(path, ...){
+  path <- c(path, as.character(unlist(list(...))))
```
krlmlr: Same.

drake.Rproj
```diff
@@ -1,21 +0,0 @@
-Version: 1.0
```
krlmlr: Did you mean to remove the RStudio project file?

R/check.R
wlandau: How about when the code analysis is done? That's the easiest place to tell whether a file name in a command is in single quotes and not in `file_input()`, `file_output()`, or `knitr_input()`. See 34eb1e02dec4a5b98dfc8962bdd539c23c3e9f48.

wlandau: Oops, that commit broke the tests. Fixed in 980e2bc8d98d42862e5cecd2df5e1e2e47c44934, I think.

R/dependencies.R
wlandau: In the `deps()` examples, I suppose `report.md` should have literal double quotes around it. Internally, I planned to have files represented with double quotes rather than single quotes. This avoids all that funny business with `parse()` dropping single quotes. See 980e2bc8d98d42862e5cecd2df5e1e2e47c44934.

R/dependencies.R
wlandau: Nope. `analyze_readd()` is just a static code analysis of a call to `readd()` (expected in `knitr` reports, but not commands or imported functions), and `args` is just the argument list of that call. `readd()` and `loadd()` are special because we can't just assume every argument is a target or dependency. For example, in `readd(some_target, path = "~/.drake")`, `drake` needs to know that `some_target` is a target and `"~/.drake"` is not.

R/dependencies.R
wlandau: Actually, I don't think we should even be calling `eval()` at all. I only did that to extract targets from symbols like `c("target1", "target2")` without the false positive `c()`. The code analysis should be totally static. `CodeDepends` to the rescue again! (cc @gmbecker.) See 7cd56a959ed2ca4c717d1fd4b2aad549a96b75bd.

R/read.R
wlandau: Single-quotes go away during parsing and formatting, so internally, I made `drake` turn all file names into double-quoted strings. I think this is now consistent with the rest of the current `drake` in this branch. But you're right in the sense that users might still expect to use single-quotes if they want to read file hashes from the cache. In 004b78e53ddc419d4577c01aab039bb28f4e939c, `drake` standardizes filenames before reading from the cache just in case users use single quotes.

R/utils.R
wlandau: Maybe a `purr` function, but `utils::modifyList()` doesn't do what we need here.

```r
> x <- list(a = 1, b = 2)
> y = list(b = 4, c = 3)
> utils::modifyList(x, y) # forgets about the `b` element of `x`
$a
[1] 1

$b
[1] 4

$c
[1] 3

> drake:::merge_lists(x, y)
$a
[1] 1

$b
[1] 2 4

$c
[1] 3
```

R/workplan.R
wlandau: Agreed on `as.character(c(...))`. See 2e13c8e12f69d1751712db671bcf3eaad5c8024c. Not sure if `fs_path()` is worth importing `fs`, especially given `drake`'s already large number of dependencies. Where would the file path be printed? I think pretty printing of the file paths would be most helpful in the workflow plan data frame, but I am not sure if that is possible.

R/workplan.R
wlandau: See 2e13c8e12f69d1751712db671bcf3eaad5c8024c.

drake.Rproj
wlandau: Nope, that was accidental. RStudio was crashing, and I was trying to figure out why. Added the project file back in cae7ce01b6981bed658d7a28ada4495619bd05bc.

R/utils.R
krlmlr: Maybe:

``` r
x <- list(a = 1, b = 2)
y <- list(b = 4, c = 3)
purrr::list_merge(x, !!! y)
#> $a
#> [1] 1
#> 
#> $b
#> [1] 2 4
#> 
#> $c
#> [1] 3
```

> Created on 2018-02-19 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

R/workplan.R
krlmlr: We don't really need to import *fs*, but maybe add a `"class"` attribute?

wlandau: Sorry, I'm having trouble understanding the motivation. Would we give the `class` attribute to the path/target name itself? Where would the printing happen?

wlandau:
Kirill, your feedback helped a lot. You hit all the most important pieces. The changes to the help files, vignettes, examples, and tests were far more straightforward, and I am less concerned about feedback there. (Of course, feel free to review them if you like.) I think we are ready to merge with master. Enjoy the new API, everyone!

R/cache_ui.R
```diff
@@ -226,7 +226,8 @@ targets_from_dots <- function(dots, list) {
   names <- vapply(dots, as.character, "")
   if (length(names) == 0L)
     names <- character()
-  .Primitive("c")(names, list) %>% unique
+  targets <- .Primitive("c")(names, list) %>% unique
```
krlmlr: Why use `.Primitive("c")` instead of `c` here? `names` seems to be an unclassed character (also the check above in lines 227-228 seems superfluous).

R/dependencies.R
```diff
@@ -177,43 +178,70 @@ nonfile_target_dependencies <- function(targets, config){
...
+  new_api_files <- c(deps$file_input, deps$file_output, deps$knitr_input) %>%
+    drake_unquote
+  warn_files <- setdiff(old_api_files, new_api_files)
+  if (!length(warn_files)){
+    return()
   }
+  warning(
+    "Files in a command declared with single-quotes:\n",
+    multiline_message(warn_files),
+    "\nThe way to declare files in drake is changing. ",
```
krlmlr: "is changing" sounds unclear to me, maybe: "Declaring files using single-quotes is deprecated. Instead, use file_input() ...."

R/dependencies.R
```diff
@@ -243,22 +271,176 @@ unwrap_function <- function(funct){
...
+    expr,
+    knitr_input = FALSE,
+    file_input = FALSE,
+    file_output = FALSE
+  ){
+    if (!length(expr)){
+      return()
+    } else if (is.function(expr)) {
+      expr <- unwrap_function(expr)
+      if (typeof(expr) != "closure"){
```
krlmlr: What are the other cases, primitives? Do we really need to `walk()` in these cases?

R/dependencies.R
```diff
@@ -243,22 +271,176 @@ unwrap_function <- function(funct){
...
+      walk(
+        body(expr),
+        knitr_input = knitr_input,
+        file_input = file_input,
+        file_output = file_output
+      )
+    } else if (is.name(expr) || is.atomic(expr)) {
+      if (knitr_input){
+        file <- declare_file(expr)
+        results$knitr_input <<- base::union(results$knitr_input, file)
```
krlmlr: It seems clearer if only `results` is updated using `<<-` and `merge_lists`, and perhaps only in one place:

```r
  if (...) {
    new_results <- ...
  } else if (...) {
    new_results <- ...
  } ...

  # Update results declared in code_dependencies()
  results <<- merge_list(results, new_results)
```

We could also think about having `walk()` return `new_results` and get rid of `<<-` altogether.

R/dependencies.R
```diff
@@ -243,22 +271,176 @@ unwrap_function <- function(funct){
...
+}
+
+analyze_loadd <- function(expr){
+  args <- parse_loadd_arg_list(expr)
+  c(unnamed_in_list(args), args[["list"]])
+}
+
+analyze_readd <- function(expr){
+  args <- parse_loadd_arg_list(expr)
+  c(unnamed_in_list(args), args[["target"]])
```
krlmlr: I think I understand why you query `target`. It feels that `match.call()` would be slightly safer to use here:

``` r
match.call(drake::readd, as.call(alist(readd, 1)))
#> readd(target = 1)
match.call(drake::readd, as.call(alist(readd, 2, target = 1)))
#> readd(target = 1, character_only = 2)
match.call(drake::loadd, as.call(alist(readd, 1, 2, 3, list = list(4))))
#> readd(1, 2, 3, list = list(4))
```

> Created on 2018-02-20 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

R/read.R
krlmlr: I wonder if `readd(file_output("report.md"))` should be the preferred syntax instead. This would finally justify the `"class"` attribute ;-) (And maybe shorten to `file_in()`, `file_out()` and `knitr_in()`? But I don't have a strong preference.)

R/utils.R
```diff
@@ -26,3 +32,25 @@ braces <- function(x) {
...
+  ) %>%
+    setNames(nm = names)
+}
+
+clean_dependency_list <- function(x){
+  if (!length(x)){
+    return(character(0))
+  }
+  x %>%
+    unlist() %>%
```
krlmlr: `unlist()` on a named list computes funny names for the intermediate object, maybe add an extra `unname()` (and an `as.character()` after the following `unname()` to get rid of the shortcut in line 48)?

R/workplan.R
```diff
@@ -127,20 +107,47 @@ drake_plan <- function(
...
   plan <- tibble(
     target = targets,
     command = commands
   )
   from_dots <- plan$target %in% names(commands_dots)
-  if (file_targets){
+  if (length(file_targets) || identical(strings_in_dots, "filenames")){
+    warning(
+      "The `file_targets` and `strings_in_target` are deprecated. ",
+      "See the help file examples of `drake_plan()` to see the new ",
```
krlmlr: Suggest to change order: "Use ..., see ?drake_plan for more info."

R/workplan.R
```diff
@@ -149,3 +156,133 @@ drake_plan_override <- function(target, field, config){
...
+  path <- c(path, as.character(c(...)))
+  if (length(path) != 1){
+    warning(
+      "In file_output(), the `path` argument must ",
+      "have length 1. Supplied length = ", length(path), ". ",
+      "using first file output: ", path[1], ".",
+      call. = FALSE
+    )
+  }
+  as.character(path[1])
```
krlmlr: This still gives wrong results if `path` is a list, e.g. `file_output(list(a = 1, b = 2))`. Can we just get rid of the `...` argument and fail early if the user tries to supply a second argument?

R/workplan.R
```diff
@@ -149,3 +156,133 @@ drake_plan_override <- function(target, field, config){
...
+#' vis_drake_graph(config)
+#' # See also `knitr_input()`. `knitr_input()` is like `file_input()`
+#' # except that it analyzes active code chunks in your `knitr`
+#' # source file and detects non-file dependencies.
+#' # That way, updates to the right dependencies trigger rebuilds
+#' # in your report.
+#' })
+#' }
+file_input <- function(...){
+  as.character(c(...))
```
krlmlr:
```r
  structure(
    as.character(c(...)),
    class = "drake_file_input"
  )
```

?

README.md
```diff
@@ -97,7 +101,7 @@ make(my_plan)
 ## target coef_regression2_small
 ## target summ_regression2_large
 ## target summ_regression2_small
-## target 'report.md'
+## target "report.md"
```
krlmlr: Maybe even show `target file_output("report.md")` here?

inst/testing/test.Rmd
```diff
@@ -49,10 +49,10 @@ g <- function(){
 function(){
   f(drake:::loadd(target18, character_only = IGNORE_THIS) + var)
 }
-readd("'file1'")
-readd(target = "'file2'")
-readd(target = "'file3'", character_only = TRUE)
-readd(target = "'file4'", character_only = FALSE)
-loadd("'file5'")
-loadd(list = "'file6'")
+readd("\"file1\"")
```
krlmlr: `readd(file_output("file1"))` ?

man/dependency_profile.Rd
```diff
@@ -31,7 +31,7 @@ load_basic_example() # Load drake's canonical exmaple.
 con <- make(my_plan) # Run the project, build the targets.
 # Get some example dependency profiles of targets.
 dependency_profile("small", config = con)
-dependency_profile("'report.md'", config = con)
+dependency_profile("\\"report.md\\"", config = con)
```
krlmlr: `dependency_profile(file_output("report.md"), config = con)` ?

R/workplan.R
```diff
@@ -10,23 +10,30 @@
...
 #'   with commands assigned to them. See the examples for details.
-#' @param list A named list of targets, where the values
-#'   are commands.
-#' @param file_targets logical, whether the targets should be
-#'   (single-quoted) external file targets.
-#' @param strings_in_dots Character scalar,
+#' @param list A named character vector of commands
+#'   with names as targets.
+#' @param file_targets deprecated argument. See [file_output()],
+#'   [file_input()], and [knitr_input()] for the new way to work
```
krlmlr: new -> current?

man/drake_plan.Rd
```diff
@@ -52,60 +63,31 @@ with a \code{target} column and a \code{command} column.
...
+make(mtcars_plan) # Makes `mtcars.csv` and then `value`
+head(readd(value))
+# You can use knitr inputs too. See the top command below.
+load_basic_example()
+head(my_plan)
+# The `knitr_input("report.Rmd")` tells `drake` to dive into the active
+# code chunks to find dependencies.
+# There, `drake` sees that `small`, `large`, and `coef_regression2_small`
+# are loaded in with calls to `loadd()` and `readd()`.
+deps("report.Rmd")
```
krlmlr: `deps(file_input("report.Rmd"))` ?

tests/testthat/test-Makefile.R
```diff
@@ -103,7 +100,7 @@ test_with_dir("basic Makefile stuff works", {
   ))
   expect_equal(stamps, stamps2)
 
-  targ <- "'intermediatefile.rds'"
+  targ <- "\"intermediatefile.rds\""
```
krlmlr: `targ <- file_output("intermediatefile.rds")` ?

vignettes/graph.Rmd
```diff
@@ -78,7 +78,7 @@ style="border: none; box-shadow: none">
 Graphs can grow enormous for serious projects, so there are multiple ways to focus on a manageable subgraph. The most brute-force way is to just pick a manual `subset` of nodes. However, with the `subset` argument, `vis_drake_graph()` may drop intermediate nodes and edges.
 
 ```{r subsetgraph}
-vis_drake_graph(config, subset = c("regression2_small", "'report.md'"))
+vis_drake_graph(config, subset = c("regression2_small", "\"report.md\""))
```
krlmlr: `..., subset = c(..., output_file("report.md")))` ?

krlmlr:
Great job, by the way! (Never mind the feedback, I'm being picky because I think it's important to get this right.)

R/cache_ui.R
wlandau: Inline is fine, all these are easy to fix. Sorry if you feel I merged too early. I expect us to continue refining this on `master`.

The `Primitive("c")`, etc. is leftover from when I was copying `base::remove()`. Changed in 23975178fa729ef1c277c816a3858f999dcd75c9.

R/dependencies.R
wlandau: You're right. See 49f7c92e22355a1497b57f8f39d4b91b86d6ac72.

R/dependencies.R
wlandau: Added a condition for primitives in 83e44b5114a2e518fb74ac886ec9ca924de3b7b5. Pairlists may be another case, but I think we've covered those already in `if(!length(expr)){...}` and `else if(is.recursive(expr)){...}`.

R/dependencies.R
wlandau: And it shortens the code a lot. I like that. See 3d420f9459116dc6f7e5ab248b075ce02c4ab038. I would be open to getting rid of `<<-`.

R/dependencies.R
wlandau: Adding `match.call()` does seem to make things safer. Added in 2bd2f0aebdeb60e0b624df150dfa0503aeaaf771.

R/utils.R
wlandau: Do you mean two `unname()`s in a row? Not exactly sure if that's what you meant. Added an `as.character()` in 2bd06c036ab780691a29b6b9db6d11a2c7415ec3.

R/workplan.R
wlandau: Does 7158210b617dd104297788468540a1d8db05b7a8 have the order you had in mind?

R/workplan.R
wlandau: I assumed `file_output(list(a = 1, b = 2))` return `"1"` (current behavior). What do you think it should return?

Your comment actually spurred me to refactor the code analysis. I will have another PR for your review soon.

R/workplan.R
wlandau: Just to assign a class to `file_input()` deps. Feel free to change.

README.md
wlandau: How about `target file "report.md"`? Will wrap into the new PR.

R/workplan.R
wlandau: Wrapping into the new PR.

inst/testing/test.Rmd
wlandau: This would justify the `"class"` attribute on `file_output()`, etc. Will submit as a new issue.

man/dependency_profile.Rd
wlandau: Same as https://github.com/ropensci/drake/pull/258#discussion_r169291609.

man/drake_plan.Rd
wlandau: Same as https://github.com/ropensci/drake/pull/258#discussion_r169291609.

tests/testthat/test-Makefile.R
wlandau: Maybe. But in this case, `file_output()` would have to add extra quotes. Seems to conflict with the other uses of `file_output()`.

vignettes/graph.Rmd
wlandau: Same as https://github.com/ropensci/drake/pull/258#discussion_r169291609.

wlandau:
Thanks, Kirill! I want to get it right too. I merged to master because I thought it was ready for beta testing, and I expect us to return to this.

By the way, I would be open to synonyms `file_in()`, `file_out()`, and `knitr_in()`.

R/read.R
wlandau: Just wrapped these synonyms into #265.

tests/testthat/test-Makefile.R
krlmlr: Not if we have a `"class"` attribute.

R/utils.R
krlmlr: `unname() %>% unlist() %>% unname()` , the second may be unnecessary but I'm not sure.

krlmlr:
Reviewing after the fact is fine for me.

wlandau:
Great! I'll go ahead and merge some on my own (like #263) and wait for you for others (like #259).

tests/testthat/test-Makefile.R
wlandau: If we can make sure the attribute is not dropped throughout the API, then it does seem like the right thing to do in the long run. Needs a bit of refactoring.

PR #260, opened 2018-02-18
Fix doc link and typo

krlmlr:
Using `@inheritParams` for the `cache`, `path` and `search` arguments, inheriting from `cached()`. Also fixes return type documentation for `read_drake_seed()`.


wlandau:
Thanks, Kirill! I appreciate all this cleanup.

PR #270, opened 2018-02-21
Master announces builds for futures

krlmlr:
The green `target` is back. The announcement now happens when the future is created, but it's better than nothing and the best we can do as long as there's no status communication.


krlmlr:
On some futures the announcement now happens twice. Will investigate.

wlandau:
@krlmlr I followed through with your idea in #275, now on `master`. Please let me know if builds still go unannounced for some `future` backends.

krlmlr:
Thanks for looking into it, I didn't think through my implementation.

PR #274, opened 2018-02-21
Tweak warnings in drake_plan()

krlmlr:
- clarify user actions
- un-deprecate `strings_in_dots` argument (we need to ask the user to pass in `"literals"` for full backward compatibility)


wlandau:
Thanks, Kirill. The docs of `drake_plan()` are much clearer now.

PR #284, opened 2018-02-26
Update expand() [deprecated] to expand_plan() in instructional commen…

bmchorse:
# Summary

This PR is a tiny update to the `basic example` vignette and R code. 

In the comments, the user encouraged to expand the analysis plan. `expand()` is suggested, but if the user does so, it throws a deprecation warning; I've updated it to `expand_plan()` so that users following the suggestion will use the correct function.


wlandau:
Thanks, Brianna! I often miss things like this. Will merge when the Travis build completes.

PR #309, opened 2018-03-09
Revert "Set use_cache = FALSE in readd()"

krlmlr:
This reverts commit de353924e0dec55d37ba5ee9ffd1e8a2adaa31d4.

I think caching here is as important as for `loadd(lazy = "bind")`.

### DIFF
--- a/R/read.R
+++ b/R/read.R
@@ -52,7 +52,7 @@ readd <- function(
   cache$get(
     standardize_filename(target),
     namespace = namespace,
-    use_cache = FALSE
+    use_cache = TRUE
   )
 }
 


wlandau:
I assume you meant "Set use_cache = TRUE" in the title? If we trust `storr` to not max out the memory (https://github.com/richfitz/storr/issues/69#issuecomment-370998999), this is fine. I guess we'll find out pretty easily if we run into memory issues in `make()`.

wlandau:
Given https://github.com/richfitz/storr/issues/69#issuecomment-372184072, can we revisit this? Maybe we should make a PR for `storr`?

krlmlr:
Let's help improve *storr*. Not sure how much this particular instance of `use_cache = TRUE` is a problem for real-world projects.?

wlandau:
I am not sure about the implications for real-world projects. The whole time I have been using `drake`, I have unknowingly used the default `use_cache = TRUE` without any apparent problems. But I have not worked with anything as massive as what @kendonB is doing.

kendonB:
I haven't noticed major memory problems but also haven't used readd in a way that would cause problems.

wlandau:
That's good to hear. Using [`prune_envir()`](https://github.com/ropensci/drake/blob/e78d2c270bde1dd3a9d238fcc523ffd0ce1b6c69/R/envir.R#L26),  `make()` actually does call `readd()` through `loadd()` when it shuffles around targets to try to be memory efficient.

PR #311, opened 2018-03-10
Change verbosity levels

### LINKED ISSUE #165: Consider integer argument to verbose?
``import ...`` clogs up the screen but I'd still like to see ``check ...``. An integer argument to verbose could achieve this. 0 = FALSE, 1=TRUE, 2=No imports.

A better solution IMO would stop printing individual imports altogether and just do one chunk with a ... per stage. Now, we already see ``check`` for the imports anyway, so I don't see a big cost in switching the import printing off by default.

krlmlr:
All existing levels are increased by one, existing code will be less chatty in general. The new level 1 only shows targ
ets.

I have reviewed changes to the website, looks good, but I didn't carry out a comprehensive test. The "drake-pitch" slides also have much less log info now.

Closes #165.

### DIFF
--- a/R/config.R
+++ b/R/config.R
@@ -49,11 +49,10 @@
 #' @param verbose logical or numeric, control printing to the console.
 #'   \describe{
 #'     \item{0 or `FALSE`:}{print nothing.}
-#'     \item{1 or `TRUE`:}{print checks and targets to build.}
-#'     \item{2:}{print checks, targets, to build,
-#'       and any potentially missing items.}
-#'     \item{3:}{full verbosity: print checks, targets to build, potentially
-#'       missing items, and imports.}
+#'     \item{1 or `TRUE`:}{print only targets to build.}
+#'     \item{2:}{in addition, print checks and cache info.}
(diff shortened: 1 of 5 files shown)


wlandau:
Thanks for your work on this, Kirill. I will try out your branch tonight. The code and docs  look right, and I think we'll be able to merge when the tests catch up.

krlmlr:
I had fixes for the tests locally, forgot to commit. What's our policy regarding NEWS entries?

wlandau:
I think we should put in NEWS anything remotely nontrivial that astute users are likely to notice. Not a hard and fast policy, though, and I know I forget way too often. I am open to other approaches.

krlmlr:
I tend to collect changelogs from commit messages to master into NEWS. This means that we don't need to think about NEWS as we go, only for (relatively infrequent) minor version bumps. My workflow is semi-automated but very convoluted, I'm planning to simplify and publish it at some point. Until then I'm happy to adapt.

wlandau:
Interesting. I think a manual `NEWS.md` is not so much of a hassle at the moment, but some streamlined automation could help eventually.

PR #314, opened 2018-03-11
Fix #303

### LINKED ISSUE #303: Incorrectly outdated targets after loadd()
Calling `loadd()` changes `outdated(drake_config(plan))`, but not `outdated(read_drake_config())`. This doesn't happen if `x` is created from a constant such as `42`.

``` r
writeLines("test", "test.txt")

pkgload::load_all("testpkg")
#> Loading testpkg
library(drake)
plan <- drake_plan(
  x = readLines(file_in("test.txt")),
  strings_in_dots = "literals"
)
make(plan)
#> cache /tmp/RtmpArc3UM/.drake
#> connect 1 import: plan
#> connect 1 target: x
#> check 2 items: file "test.txt", readLines
#> check 1 item: x
#> target x
outdated(read_drake_config())
#> cache /tmp/RtmpArc3UM/.drake
#> check 2 items: file "test.txt", readLines
#> unload 1 item: x
#> check 1 item: x
#> character(0)
outdated(drake_config(plan))
#> cache /tmp/RtmpArc3UM/.drake
#> connect 1 import: plan
#> connect 1 target: x
#> check 2 items: file "test.txt", readLines
#> check 1 item: x
#> character(0)
loadd()
#> cache /tmp/RtmpArc3UM/.drake
outdated(read_drake_config())
#> cache /tmp/RtmpArc3UM/.drake
#> check 2 items: file "test.txt", readLines
#> unload 1 item: x
#> check 1 item: x
#> character(0)
outdated(drake_config(plan))
#> cache /tmp/RtmpArc3UM/.drake
#> connect 10 imports: close, file "testpkg/DESCRIPTION", UseMethod, is.cha...
#> connect 1 target: x
#> check 5 items: .Internal, file "test.txt", is.character, on.exit, UseMethod
#> check 2 items: close, file
#> check 1 item: readLines
#> check 1 item: x
#> [1] "x"
rm(x)
#> Warning in rm(x): object 'x' not found
outdated(read_drake_config())
#> cache /tmp/RtmpArc3UM/.drake
#> check 2 items: file "test.txt", readLines
#> check 1 item: x
#> character(0)
outdated(drake_config(plan))
#> cache /tmp/RtmpArc3UM/.drake
#> connect 10 imports: close, file "testpkg/DESCRIPTION", UseMethod, is.cha...
#> connect 1 target: x
#> check 5 items: .Internal, file "test.txt", is.character, on.exit, UseMethod
#> check 2 items: close, file
#> check 1 item: readLines
#> check 1 item: x
#> [1] "x"
make(plan)
#> cache /tmp/RtmpArc3UM/.drake
#> connect 10 imports: close, file "testpkg/DESCRIPTION", UseMethod, is.cha...
#> connect 1 target: x
#> check 5 items: .Internal, file "test.txt", is.character, on.exit, UseMethod
#> check 2 items: close, file
#> check 1 item: readLines
#> check 1 item: x
#> target x
outdated(read_drake_config())
#> cache /tmp/RtmpArc3UM/.drake
#> check 5 items: .Internal, file "test.txt", is.character, on.exit, UseMethod
#> unload 1 item: x
#> check 2 items: close, file
#> check 1 item: readLines
#> check 1 item: x
#> character(0)
```

> Created on 2018-03-06 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

wlandau:
# Summary

- In `loadd()`, exclude foreign imports: i.e., imports that were not explicitly taken from `envir`. Foreign imports can come from a package or an ancestor of `envir`.
- in `flexible_get()`, don't just `get(target)`. Use `get(target, envir)` to be sure that `flexible_get()` searches for `target` in `envir`, an ancestor of `envir`, or possibly a package. That way, `drake` does not creep into the global environment uninvited. This change could invalidate targets for strangely-configured workflows, and most users are not likely to notice it.

# GitHub issues fixed

- Ref: #303

# Checklist

- [x] This pull request is ready for review and/or merge.

### DIFF
--- a/R/meta.R
+++ b/R/meta.R
@@ -64,7 +64,7 @@ drake_meta <- function(target, config = drake::read_drake_config()) {
   meta <- list(
     target = target,
     imported = !(target %in% config$plan$target),
-    foreign = !(target %in% ls(config$envir)),
+    foreign = !(exists(x = target, envir = config$envir, inherits = FALSE)),
     missing = !target_exists(target = target, config = config),
     seed = seed_from_object(list(seed = config$seed, target = target))
   )
--- a/R/meta.R
+++ b/R/meta.R
@@ -64,6 +64,7 @@ drake_meta <- function(target, config = drake::read_drake_config()) {
   meta <- list(
     target = target,
     imported = !(target %in% config$plan$target),
+    foreign = !(target %in% ls(config$envir)),
     missing = !target_exists(target = target, config = config),
     seed = seed_from_object(list(seed = config$seed, target = target))
   )
--- a/R/read.R
+++ b/R/read.R
@@ -234,6 +234,9 @@ exclude_foreign_imports <- function(targets, cache, jobs){
 }
 
 is_not_foreign_import_obj <- function(target, cache){
+  if (is_file(target)){
+    return(TRUE)
+  }
   if (!cache$exists(key = target, namespace = "meta")){
     return(FALSE)
   }
@@ -243,9 +246,7 @@ is_not_foreign_import_obj <- function(target, cache){
     character_only = TRUE,
     verbose = FALSE
   )
-  identical(meta$imported, FALSE) ||
-    identical(meta$foreign, FALSE) ||
-    is_file(target)
+  identical(meta$imported, FALSE) || identical(meta$foreign, FALSE)
 }
 
 parse_lazy_arg <- function(lazy){
--- a/R/read.R
+++ b/R/read.R
@@ -61,7 +61,11 @@ readd <- function(
 #' current workspace (or environment `envir` if given). Defaults
 #' to loading the entire cache if you do not supply anything
 #' to arguments `...` or `list`.
-#' @seealso [cached()], [built()],
+#' @details `loadd()` excludes foreign imports:
+#'   R objects not originally defined in `envir`
+#'   when [make()] last imported them.
+#'   To get these objects, use [readd()].
+#' @seealso [readd()], [cached()], [built()],
 #'   [imported()], [drake_plan()], [make()],
 #' @export
 #' @return `NULL`
@@ -141,6 +145,9 @@ readd <- function(
 #' loadd(coef_regression2_small, deps = TRUE)
 #' ls()
 #' # Load all the imported objects/functions.
+#' # Note: loadd() excludes foreign imports
+#' # (R objects not originally defined in `envir`
+#' # when `make()` last imported them).
 #' loadd(imported_only = TRUE)
 #' ls()
 #' # Load all the targets listed in the workflow plan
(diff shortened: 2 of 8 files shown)


R/meta.R
```diff
@@ -64,6 +64,7 @@ drake_meta <- function(target, config = drake::read_drake_config()) {
   meta <- list(
     target = target,
     imported = !(target %in% config$plan$target),
+    foreign = !(target %in% ls(config$envir)),
```
krlmlr: `exists()` might be faster here.

R/read.R
```diff
@@ -214,6 +224,30 @@ loadd <- function(
...
+  }
+  meta <- diagnose(
+    target = target,
+    cache = cache,
+    character_only = TRUE,
+    verbose = FALSE
+  )
+  identical(meta$imported, FALSE) ||
+    identical(meta$foreign, FALSE) ||
+    is_file(target)
```
krlmlr: Do we need to call `diagnose()` if `is_file(target)` ?

R/read.R
```diff
@@ -206,6 +211,11 @@ loadd <- function(
   if (!replace){
     targets <- setdiff(targets, ls(envir, all.names = TRUE))
   }
+  targets <- exclude_foreign_imports(
```
krlmlr: The changes in this function conceptually seem to fit better in `drake_select()`. Do we really need `build_times()` or `clean()` for imported targets (these are other callers of `drake_select()`)? Does that function have access to all relevant information?

R/meta.R
wlandau: I tried `exists()` at first, but it also includes the ancestors of `envir`, so I was getting bad results. But you're right in the sense that slow set operations are too widespread throughout the code. I do want to change that at some point.

R/read.R
wlandau: Imported files do not have the problem in #303, so that line includes them. The `is_file(target)` makes the new behavior agree with the old behavior when possible. I do think we should think about excluding files in general from `loadd()`, but I think it is out of scope of this PR.

R/read.R
wlandau: To answer both this question and your first question: the only reason to worry about foreign imports is #303, where we risk putting a function in an environment where it does not belong. The only API function with this problem is `loadd()`.

Another thing: I really think we should screen for foreign imports towards the end of `loadd()` rather than in `drake_select()` because that's the only place we have the actual list of targets. For example, if the `deps` argument is `TRUE`, we are actually loading the *dependencies* of the items enumerated with `drake_select()`.

R/meta.R
krlmlr: `exists(inherits = FALSE)` perhaps?

R/read.R
krlmlr: Can we check `is_file()` without invoking `diagnose()`?

wlandau: Yes: 2d1180b3c214922689aff1d636e8d871390c00b4.

R/meta.R
wlandau: Yes: 2d1180b3c214922689aff1d636e8d871390c00b4.

wlandau:
Seems like we can merge now and make minor adjustments going forward. Thanks for the feedback, as always.

Issue #299, opened 2018-03-06
Optionally ignore parts of commands


wlandau:
### Problem

```r
plan <- drake_plan(
  datasets = mclapply(1:8, get_dataset, mc.cores = 8),
  analyses = lapply(datasets, analyze_dataset),
  summaries = lapply(analyses, summarize_analysis)
)
```

What if we want to change `mc.cores`? The content of `datasets` won't change, but `drake` will think `datasets` is out of date.

### Solution: a new `ignore()` function

```r
plan <- drake_plan(
  datasets = mclapply(1:8, get_dataset, mc.cores = ignore(8)),
  analyses = lapply(datasets, analyze_dataset),
  summaries = lapply(analyses, summarize_analysis)
)
```

`Drake` should not react when 8 is changed to something else. We can make this happen by modifying [`standardize_command()`](https://github.com/ropensci/drake/blob/e87f05ad70a9c605df885c07df1bc5fa484db352/R/commands.R#L64).  Also, `drake` should not reproducibly track any symbols mentioned inside `ignore()`, so the code analysis needs to change.

krlmlr:
I think this is orthogonal to within-target parallelism, but a very useful feature anyway. Related: An `always()` function that will make sure the code is always run.

Can we name these functions `drake_isolate()` and `drake_xxx()` (in analogy to Shiny, perhaps with `xxx` = `always`)?

wlandau-lilly:
What would an `always()` function do that is different from the `"always"` trigger?

We could consider `drake_isolate()`, but I think `isolate()` is most similar to the `file_in()`, `file_out()`, and `knitr_in()` functions, none of which has a `drake_` prefix.

krlmlr:
`always()` would work pretty much the same as an `"always"` trigger, but you would be able to express it in a `drake_plan()`.

`isolate()` will collide with *shiny*, not sure if we need to care about that.

wlandau:
Good point on `isolate()`. I still like the word `ignore()`, and I am leaning away from `always()` because it puts us in a situation where changing the trigger invalidates the target. We would also have to handle the case where `always()` appears in a command and the explicit entry in the `trigger` column is something else.

krlmlr:
Can we prune the `always()` call during code analysis? I'd think `always()` either in the command or in the `trigger` column wins.

How do we create the `trigger` column in `drake_plan()`

wlandau:
> Can we prune the always() call during code analysis?

Pruning `always()` would cover it.

[...]

PR #317, opened 2018-03-11
Implement `ignore()`

wlandau:
# Summary

In commands in the workflow plan, you can ignore(`some_code`) to

1. Force `drake` to not track dependencies in `some_code`, and
2. Ignore any changes in `some_code` when it comes to deciding which target are out of date.

# Help requested

`Drake` uses `standardize_command()` to create a reproducibly-tracked representation (kernel) of a command. Changes to the kernel invalidate the target. As part of this PR, code inside `ignore()` is removed from the kernel. I want to use proper code analysis rather than text analysis to to accomplish this, but I do not know how. The failing tests expose problems with the text analysis approach.

# GitHub issues fixed

- Ref: #299 

# Checklist

- [x] I have added [`testthat`](https://github.com/r-lib/testthat) unit tests to [`tests/testthat`](https://github.com/ropensci/drake/tree/master/tests/testthat) to test new functionality.
- [x] Tests pass locally locally with `devtools::check()`
- [x] Ready to merge.


R/workplan.R
```diff
@@ -260,7 +260,7 @@ file_in <- function(...){
 #'   one file output per command. See the examples
 #'   for a full explanation.
 #' @export
-#' @seealso file_in knitr_in
+#' @seealso [file_in()], [knitr_in(()], [ignore()]
```
krlmlr: Two opening parens instead of one.

wlandau: Thanks: 379e5c7c377e7f6538d0073e89b7a389b7b5c5a2

wlandau:
> I think standardize_command() could be implemented much in a same way like code_dependencies(): Walk the AST, return the input, unless it's an ignore() call.

I would like to, but if we [return the input](https://github.com/ropensci/drake/blob/e78d2c270bde1dd3a9d238fcc523ffd0ce1b6c69/R/dependencies.R#L338), would we risk getting a nested list instead of an expression or language object? I think that's the final piece of this PR/issue.

krlmlr:
For walking and updating the AST and updating, assigning via `[<-` seems to preserve type:

``` r
cl_orig <- call("fun", quote(arg1), quote(arg2))
cl <- cl_orig
cl
#> fun(arg1, arg2)
cl[] <- cl[]
cl
#> fun(arg1, arg2)
identical(cl_orig, cl)
#> [1] TRUE
```

> Created on 2018-03-11 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

wlandau:
I'm not sure how `cl[] <- cl[]` will help us recurse on the elements of a language object, so I [put in a `for` loop](https://github.com/ropensci/drake/blob/59baef7504d3f077ee59a879fcccf295709b14d9/R/commands.R#L100) for now. Willing to discuss and improve. At least now we're using proper code analysis to walk the AST. And now `ignore()` appears to work not only for commands but also for imported functions.

R/commands.R
```diff
@@ -75,18 +76,47 @@ standardize_command <- function(x) {
...
+    body(x) <- recurse_ignore(body(x))
+  } else if (
+    length(x) > 0 &&
+    is.language(x) &&
+    (is.call(x) || is.recursive(x))
+  ){
+    if (is_ignore_call(x)) {
+      x <- quote(ignore())
+    } else {
+      for (i in seq_along(x)){
```
krlmlr: `x[] <- map(x, recurse_ignore)` ?

R/dependencies.R
```diff
@@ -317,7 +317,7 @@ code_dependencies <- function(expr){
       new_globals <- setdiff(
         x = wide_deparse(expr), y = drake_fn_patterns)
       results$globals <<- c(results$globals, new_globals)
-    } else if (is.call(expr) || is.recursive(expr)) {
+    } else if (is.language(expr) && (is.call(expr) || is.recursive(expr))) {
```
krlmlr: Can we extract this check to a function, maybe `is_callish()` ? Why do we need to check `is.recursive()` ?

R/dependencies.R
```diff
@@ -459,3 +461,11 @@ is_loadd_call <- function(expr){
 is_readd_call <- function(expr){
   wide_deparse(expr[[1]]) %in% readd_fns
 }
+
+is_ignore_call <- function(expr){
+  wide_deparse(expr[[1]]) %in% ignore_fns
+}
+
+ignore_pattern <- paste0(
```
krlmlr: Do we still need this?

R/commands.R
wlandau: `purrr::map(x, recurse_ignore)` and purrr::map(x[], recurse_ignore)` both give me "Error: `.x` is not a vector (language)".

R/dependencies.R
wlandau: Nope. Removed in b8d263693e0bddb80a06fa479e448a6e235ee856.

R/dependencies.R
wlandau: `is_callish()`, yes. See b8d263693e0bddb80a06fa479e448a6e235ee856
. For both `code_dependencies()` and `ignore_ignore()`, I remember thinking we don't need `is.recursive()` at first, but it in my experience it turns out to be needed for some edge cases: for example, diving into explicit `{` blocks.

R/commands.R
krlmlr: `map(as.list(x))` maybe?

wlandau: Maybe, but then it seems like we would be changing the type again.

krlmlr: Also with `x[] <- map(as.list(x), ...)` ?

``` r
x <- call("fun", quote(a1), quote(a2))
x
#> fun(a1, a2)
x[] <- purrr::map(as.list(x), ~as.name(paste0("x_", .)))
x
#> x_fun(x_a1, x_a2)
```

> Created on 2018-03-16 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

wlandau: `x[] <- purrr::map(as.list(x), recurse_ignore)` works. Thanks, Kirill! Fixed in 3775026f1c899769a9bbcd9014f294cefe7595b5.

PR #318, opened 2018-03-11
Move GitHub Pages to docs

### LINKED ISSUE #312: Serve GitHub Pages from docs directory
I've been struggling to accept this as a better solution myself for a long time, because a separate documentation branch seemed neat. I see the following advantages from [serving from `/docs`](https://help.github.com/articles/configuring-a-publishing-source-for-github-pages/#publishing-your-github-pages-site-from-a-docs-folder-on-your-master-branch):

- we don't need rawgit anymore, and we don't need to update hashes because we can link directly to the artifact on the GitHub Pages  (this alone feels worth it)
- we can preview pkgdown changes in pull requests if necessary
- docs and source are better synchronized

The project is bigger when cloning or installing from GitHub, but we already have `/images`, and I don't think `/docs` adds that much more.

krlmlr:
Also reorganized scripts, `docs.R` now calls `images.R` which lives in the root. The first commit is import-only, look at the other commits for review.

See https://help.github.com/articles/configuring-a-publishing-source-for-github-pages/ for changing the source of the pages. The `gh-pages` branch can be deleted, but it also can stay around for some time as backup in case we need to revert.

Closes #312.


wlandau:
FYI: c6576e8a08e93d0703af4ce28489856529d6bda4 has minor tweaks, including an `images` folder just for the manually-created images like `logo.svg` (all copied to `docs/images/` for hosting). That way, all of `docs` is auto-generated.

wlandau:
I just checked all the rendered vignettes and the `pkgdown` site articles, and things look good.

krlmlr:
We could use the *tic* package for deployment, I have set up the *DBI* package and some others using this package, and I can set up *drake* too. I'm slightly worried about the variation from build times and paths, we're going to get quite a lot of noise from spurious changes.

wlandau:
Sounds super convenient. But I agree, there will be some noise. Maybe we should not always rebuild the images. With the reduced verbosity from #311, paths should be less of a problem.

krlmlr:
Let's get acquainted with the new workflow before setting up automation.

PR #321, opened 2018-03-14
Fix #320: ignore the `magrittr` dot (`.`)

### LINKED ISSUE #320: magrittr dot (.) always shows as a missing import
whenever I use magrittr pipe `%>%` to chain functions in a command, if I explicitly use the dot argument, either for clarity or because it is necessary, drake always shows `.` as a missing import.

Exampl;e commands where this has happened:
```r
raw_data = read_csv("data.csv") %>% mutate(.data = ., foo = bar + baz)
complete_data = raw_data %>% filter(complete.cases(.))
model = complete_data %>% lm("x ~ y", data = .)
```

# suggested action
Since `drake` imports the pipe anyway, it should be on the lookout for the dot argument, and should recognize that it is a fictitious target. Any command which uses this will already have the real dependency already, so we can safely ignore the dot.

wlandau:
# Summary

From #320 and https://github.com/duncantl/CodeDepends/issues/26, the `magrittr` dot (`.`) is sometimes treated as a dependency even when it is just a placeholder symbol. Detection depends on the context of the rest of the code in the static analysis. In the case of `drake`, this is a bit mysterious, and the behavior on `master` may invalidate targets unpredictably.

This PR forces (`.`) to never be a dependency, even when the user references it outside the intended tidyverse context. So if the user tries to reference (`.`) in a command as an ordinary variable, `drake` will still ignore it. That is fine with me because it is almost never something we should expect users to do and this PR documents it in the `caution.Rmd vignette. I think we should maximize consistency and minimize surprises.

@AlexAxthelm, I wanted to add you as a reviewer for this PR, but I cannot request anyone outside the list. May I add you as a collaborator in the repo settings for subsequent PRs?

# GitHub issues fixed

- Ref: #320

# Checklist

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -357,14 +357,14 @@ find_globals <- function(expr){
   }
   # Warning: In collector$results(reset = reset) :
   #  partial argument match of 'reset' to 'resetState'
-
   suppressWarnings(inputs <- CodeDepends::getInputs(expr))
   base::union(
     inputs@inputs,
     names(inputs@functions)
   ) %>%
     setdiff(y = c(formals, drake_fn_patterns)) %>%
-    Filter(f = is_parsable)
+    Filter(f = is_parsable) %>%
+    setdiff(y = ".") # Exclude the magrittr dot
 }
 
 analyze_loadd <- function(expr){
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # Version 5.1.0
 
+- Forcibly exclude the dot (`.`) from being a dependency of any target or import. This enforces more consistent behavior in the face of the current static code analysis funcionality, which sometimes detects `.` and sometimes does not.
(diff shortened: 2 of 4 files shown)


wlandau:
Thanks, Alex! Kirill, free to chime in when it is convenient for you.

R/dependencies.R
```diff
@@ -357,14 +357,14 @@ find_globals <- function(expr){
   }
   # Warning: In collector$results(reset = reset) :
   #  partial argument match of 'reset' to 'resetState'
-
   suppressWarnings(inputs <- CodeDepends::getInputs(expr))
   base::union(
     inputs@inputs,
     names(inputs@functions)
   ) %>%
     setdiff(y = c(formals, drake_fn_patterns)) %>%
```
krlmlr: Maybe just add the dot up there?

wlandau: Good catch. See 3ef462e45a3c5f10d1e8d14533ad73fd63e56f18.

Issue #283, opened 2018-02-26
Allow multiple output files for each command


wlandau:
Contemplated in #232 and #257. I just saw that [`snakemake`](https://bitbucket.org/snakemake/snakemake/) allows any number of output files, and I think `drake` should try to catch up.

```r
drake_plan(
  data = get_data(),
  tmp = {
    saveRDS(data, file_out("data.rds"))
    saveRDS(summary(data), file_out("summary.rds"))
  }
  my_plot = make_plot(file_in("data.rds", "summary.rds"))
)
```

From #232, it is not enough to just hack at the problem. I think it will take a rigorous design specification to plan clearly beforehand.

We may need to include a transformed version of the network graph. The current graph links together all the imported and generated objects/files. For the desired behavior, I think we need to collapse that into a dependency graph of job IDs. I was resisting this in #232.

krlmlr:
For R objects, the analogy is unpacking a named list: A command that returns multiple objects, of which separate components can be used by downstream targets.

wlandau:
I agree that we will need to unpack all the outputs if we have multiple. Could #285 will make it easier? `Drake` could handle a dependency graph of objects and files, while the external scheduler could have its own dependency graph of jobs. The reason I bring up different graphs again is that the scheduler's graph may be different anyway when we start thinking about multidimensional knapsack and transient workers with multiple jobs each.

krlmlr:
Maybe we want to switch to bipartite graphs internally, with two kinds of nodes: commands and targets? That would support M:N relationships between targets. The scheduler would operate on a simplified version where the target nodes are collapsed.

The other situation where the schedule graph is different from the dependency graph is in the context of a DSL that supports delayed expansion: A node/edge in the dependency graph can correspond to many nodes/edges in the schedule graph.

Does the terminology make sense: "schedule graph" vs. "dependency graph"?

krlmlr:
I also feel that it would pay off to consult a text book about scheduling first. Just to make sure we don't reinvent too many wheels. I'm not familiar with the literature, though.

wlandau:
Re: graphs: yes, I understand and support both the concept and the choice of terminology. Incidentally, `drake` is already (clumsily) using two graphs anyway: `config$graph` and `config$execution_graph`. The former is a dependency graph that does not get modified, and the latter is a schedule graph, copied from `config$graph` and trimmed over the course of staged parallelism.

[...]

PR #350, opened 2018-04-03
Multiple output files per target

wlandau:
# Summary

This PR is not ready to merge, and it may not even be the right way to attack the problem. 

Internally, this version of `drake` creates two DAGs: a dependency graph (`config$graph`) and a schedule of jobs (`config$schedule`). It uses the distinction to allow multiple file outputs per job. Any file outputs are cached and stored separately from the command's return value (the true target).

# GitHub issues addressed

- Ref: #283 (not solved)

# Checklist

- [ ] I have listed any substantial changes in the [development news](https://github.com/ropensci/drake/blob/master/NEWS.md).
- [ ] I have added [`testthat`](https://github.com/r-lib/testthat) unit tests to [`tests/testthat`](https://github.com/ropensci/drake/tree/master/tests/testthat) to confirm that any new features or functionality work correctly.
- [ ] I have tested this pull request locally with `devtools::check()`

### DIFF
--- a/R/Makefile.R
+++ b/R/Makefile.R
@@ -38,11 +38,11 @@ makefile_head <- function(config){
 }
 
 makefile_rules <- function(config){
-  targets <- intersect(config$plan$target, V(config$graph)$name)
+  config$graph <- targets_graph(config$schedule)
+  targets <- V(config$graph)$name
   cache_path <- cache_path(config$cache)
   for (target in targets){
     deps <- dependencies(target, config) %>%
-      intersect(y = config$plan$target) %>%
       time_stamp_target(config = config)
     breaker <- ifelse(length(deps), " \\\n", "\n")
     cat(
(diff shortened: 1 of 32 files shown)


krlmlr:
This PR is rather large. Are there any refactorings we could merge beforehand, to simplify review?

wlandau:
Yes, the PR is large. Bijectivity is deeply entrenched in `drake`, and breaking it required a whole lot of refactoring. However, I did try to plan and develop carefully. I am not sure if there are pieces we can merge on their own, but it is a good idea. We might be able to construct `config$schedule` without adopting it or modifying anything else. On the other hand, for `config$schedule` to even make sense, we need to stop forcing `file_out()` files to be target names, which breaks most of the tests.

The changes in `build.R`, `graph.R`, `meta.R`, `store.R`, and `triggers.R` get at the heart of the idea. To be clear, this PR is far from ready to merge. The tests and visualizations in this branch are in shambles.

wlandau:
When I attempt to solve #350 again, it will be after #440 is fully implemented. The solution will be much cleaner and more natural. It will not require 2 graphs, and it will rely less on the single existing `config$graph`.

PR #378, opened 2018-05-05
Trivial: add space in warning message

krlmlr:
Not sure if this should be an `\n` instead.


wlandau:
Either way, it's better than no whitespace at all. Thanks, Kirill.

Issue #369, opened 2018-05-01
A quicker end to staged parallelism


wlandau:
## The problem

With the exception of `make(parallelism = "future")` and `make(parallelism = "Makefile")`, `drake` has an embarrassing approach to parallelism, both literally and figuratively. It divides the dependency graph into conditionally independent stages and then runs those stages sequentially. Targets only run in parallel within those stages. In other words, `drake` waits for the slowest target to finish before moving to the next stage.

![capture](https://user-images.githubusercontent.com/1580860/39472822-a6d211b6-4d19-11e8-820c-1e4c05f637cf.PNG)

At the time, I could not find an R package to traverse an arbitrary network of jobs in parallel. (**Stop me right here if such a tool actually exists!**) Parallel computing tools like `foreach`, `parLapply`, `mclapply`, and `future_lapply` all assume the workers can act independently. So I thought staged parallelism was natural. However, I was thinking within the constraints of the available technology, and parallel efficiency suffered.

## Long-term work

In late February, I started the [`workers`](https://github.com/wlandau/workers) package. If you give it a job network and a parallel computing technology, the goal is to traverse the graph using a state-of-the-art algorithm. It has a proof-of-concept implementation and an initial [design specification](https://github.com/wlandau/workers/tree/master/specification), but it has a long way to go. At this stage, it really needs an R-focused [message queue](http://queues.io/). I was hoping [`liteq`](https://github.com/r-lib/liteq) would work, but due to an [apparent race condition](https://github.com/r-lib/liteq/issues/17) and some speed/overhead issues on my end, I am stuck for the time being.

## Now

I say we rework the `make(parallelism = "mclapply")` and `make(parallelism = "parLapply")` backends natively. As I see it, `make(parallelism = "mclapply", jobs = 2)` should fork 3 processes: 2 workers and 1 master. The master sends jobs to the workers in the correct order at the correct time, and the workers keep coming back for more without stopping. The goal is to minimize idleness and overhead.

The implementation will be similar to `make(parallelism = "future", caching = "worker")` (see [`future.R`](https://github.com/ropensci/drake/blob/master/R/future.R)). I think the implementation should go into `drake` itself because 

1. The code should be relatively short, and
2. I do not want to lock in the design of [`workers`](https://github.com/wlandau/workers) until I get the chance to do an extensive lit review of parallel graph algorithms.

## To-do's:

- [x] (Re)implement the low-overhead parallel backends.
[...]

PR #380, opened 2018-05-05
Use a different storr namespace for each status

wlandau:
# Summary

Here, I implement @krlmlr's suggestion to create a `storr` namespace for each persistent worker (ref: https://github.com/ropensci/drake/issues/369#issuecomment-386835371). There are no speed improvements yet, but this approach to the cache is much better.

# GitHub issues addressed

- Ref: #369 (not done yet)

# Checklist

### DIFF
--- a/R/mclapply.R
+++ b/R/mclapply.R
@@ -56,18 +56,17 @@ mc_master <- function(config){
   on.exit(mc_set_done_all(config))
   config$queue <- new_target_queue(config = config)
   while (mc_work_remains(config)){
-    for (worker in config$workers){
-      if (mc_is_idle(worker = worker, config = config)){
-        mc_conclude_target(worker = worker, config = config)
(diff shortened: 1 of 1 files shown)


wlandau:
Before, getting/setting worker status meant reading/writing the contents of files. Now, getting status means checking if a file exists, and setting status means removing one file and creating another. I am not sure which is faster, and it probably depends on the user's file system. However, this design seems a lot safer when it comes to potential race conditions I may have missed. The code is also a little cleaner, and it has the potential to be faster. So even though speed appears just a hair worse for @krlmlr's example (we now have `mclapply_staged` parallelism now for that) I will merge.

wlandau:
I so apparently, processes started hanging unpredictably in my tests. I think it has something to do with this PR, so I reverted it. The code is still in the `workers_namespaces` branch, but master no longer has it.

Issue #404, opened 2018-06-06
Race condition in storr causes strange red flags (not acual errors) in the long testing workflow


wlandau:
## TL;DR

I think this is all due to https://github.com/richfitz/storr/issues/80. 

## Overview

I am seeing strange results from `drake`'s long tests. I am struggling to find the root causes, and I have not been able to create a reproducible example for any of the elusive problems below. If anyone else has encountered similar issues, I would be super grateful for advice. I am delaying the next CRAN submission to see if we can resolve some of these lurking mysteries.

## The good news

The long tests seem to be passing. Except for the odd edge case on Windows (such as broken parallel socket connections) `drake` is delivering the output that the unit tests expect. And with the exception of https://github.com/mllg/batchtools/issues/197, my own real-world projects run just fine with development `drake`.

### The long tests

All of `drake`'s officially supported modes of parallel computing need to function properly, and not just in the quick tests that run on CRAN. That's why I implemented the internal [`test_scenarios()`](https://github.com/ropensci/drake/blob/fca3c5150833746d6c191ccf80a17ce0e7fa0407/R/test_scenarios.R#L85) function, which runs all the [unit tests](https://github.com/ropensci/drake/tree/master/tests/testthat) under each of [these testing scenarios](https://github.com/ropensci/drake/blob/master/inst/testing/scenarios.csv). [This file](https://github.com/ropensci/drake/blob/master/tests/scenarios/all.R) loops through all those scenarios, and I run it on Linux and Windows before each CRAN submission. I also test all the modes of parallel computing on some real-world projects I keep in-house.

### Temporary directories for testing

Every test runs inside a call to [`test_with_dir()`](https://github.com/ropensci/drake/blob/fca3c5150833746d6c191ccf80a17ce0e7fa0407/R/test.R#L62) rather than just `test_that()`. That way, the tests run in their own temporary directories so we do not need to worry about cleaning up file output.

### sh: 0: getcwd() failed: No such file or directory

This one gets me on Linux systems. Apparently, it happens when one tries to execute a command from a non-existent directory.

```r
> withr::with_options(
+   list(drake_no_processx = TRUE),
+   test_scenarios("local_future_lapply_2_future")
+ )
local_future_lapply_2_future::multisession: run
local_future_lapply_2_future::multisession: cache: ............................................................................................................................
local_future_lapply_2_future::multisession: command changes: ........
local_future_lapply_2_future::multisession: console: ................................................
[...]

PR #406, opened 2018-06-07
Make interprecess communication safer: compensate for storr issue 80

wlandau:
# Summary

From https://github.com/richfitz/storr/issues/80, I found out that `storr` trips over itself when we try to write the exact same data from different processes. Fortunately, this is not a big deal if we are writing ordinary data, but it does not bode entirely well for interprocess communication. Solutions:

- Use `$cache$duplicate()` instead of `$cache$set()` to set worker statuses. Rather than set new data, we just get the hash from a pool of common pre-set values.
- Use rigorous file locking from the [`filelock` package](https://github.com/r-lib/filelock) when we set statuses. Until we have an appropriate message queue, we cannot avoid the master and workers writing to the same worker status files.

# Related GitHub issues

This PR is a more complete solution to #404.

# Checklist


wlandau:
This is as far as I think we will get without a message queue.

PR #437, opened 2018-06-29
fix bug in future_lapply where it takes jobs from config

kendonB:
# Summary

`run_future_lapply` would incorrectly take the number of workers from `config$jobs`. I modified the worker setup function as well as the call to `future_lapply`

# Related GitHub issues

- Ref: None

# Checklist

- [X] I have read the [guidelines for contributing](https://github.com/ropensci/drake/blob/master/CONTRIBUTING.md).
- [NA] I have listed any substantial changes in the [development news](https://github.com/ropensci/drake/blob/master/NEWS.md).
- [NA] I have added [`testthat`](https://github.com/r-lib/testthat) unit tests to [`tests/testthat`](https://github.com/ropensci/drake/tree/master/tests/testthat) to confirm that any new features or functionality work correctly.
- [Travis did] I have tested this pull request locally with `devtools::check()`

### DIFF
--- a/R/future_lapply.R
+++ b/R/future_lapply.R
@@ -1,6 +1,6 @@
 run_future_lapply <- function(config){
   prepare_distributed(config = config)
-  mc_init_worker_cache(config)
+  mc_init_worker_cache(config, jobs = future::nbrOfWorkers())
   console_persistent_workers(config)
   path <- normalizePath(config$cache_path, winslash = "/")
   rscript <- grep(
--- a/R/future_lapply.R
+++ b/R/future_lapply.R
@@ -14,7 +14,7 @@ run_future_lapply <- function(config){
     wait = FALSE
   )
   future.apply::future_lapply(
-    X = mc_worker_id(seq_len(config$jobs)),
+    X = mc_worker_id(seq_len(future::nbrOfWorkers)),
     FUN = fl_worker,
     cache_path = config$cache$driver$path,
     future.globals = FALSE
--- a/R/future_lapply.R
+++ b/R/future_lapply.R
@@ -14,7 +14,7 @@ run_future_lapply <- function(config){
     wait = FALSE
   )
   future.apply::future_lapply(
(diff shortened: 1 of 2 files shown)


kendonB:
Looks like this might be an issue here as well:

https://github.com/ropensci/drake/blob/10151708a85acd52a74a0efdbbaf55697ea2fb0d/R/future.R

wlandau:
I would actually prefer to use `config$jobs` for this. It is more consistent with the rest of `drake`'s API. See also https://github.com/ropensci/drake/pull/438#issuecomment-401360610.

For what it's worth, you can now set different numbers of jobs for `imports` and `target`: `make(jobs = c(imports = 2, targets = 4))`. I hope this covers what you need.

kendonB:
make docs need updating in light of your last comment:
> For "future_lapply" parallelism, jobs only applies to the imports. To set the max number of jobs for "future_lapply" parallelism, set the workers argument where it exists: for example, call future::plan(multisession(workers = 4)), then call make(your_plan, parallelism = "future_lapply"). You might also try options(mc.cores = jobs), or see future::.options for environment variables that set the max number of jobs.

 I also strongly urge you to defer to nbrOfWorkers when using future to make targets. This is similar to overriding the `future::plan` in my mind, which is [bad form](https://www.rdocumentation.org/packages/future/versions/1.8.1/topics/plan):
> Please refrain from modifying the future strategy inside your packages / functions, i.e. do not call plan() in your code. 

It also agrees with a past version of yourself which intended to defer to `nbrOfWorkers` and have jobs just apply to imports.

wlandau:
> make docs need updating in light of your last comment

I removed the comment in 871439bbba419b6308984173e64a5e732f550c97.

> I also strongly urge you to defer to nbrOfWorkers when using future to make targets. This is similar to overriding the future::plan in my mind, which is bad form:

It would be bad form, but I believe inconsistency in `jobs` is even worse. Here, `future` is the back-end and `drake` is the front-end. I wish we could obey best practices for both, but I still believe the current behavior is the path of least friction for users.

Issue #168, opened 2017-11-23
Search branch until finding something out of date in parallel


kendonB:
I just ran into a case where a potential parallelisation wasn't realised.

The dependency structure was something like:

A -> B -> D -> E
A -> C ->          F

A, B, C, and D are up to date. Drake checked A, BC, DF, built F, then checked and built E.

I think the most efficient behavior would be:
A, B, C, and D are up to date. Drake checks A, BC, DF, E, then builds EF in parallel.

wlandau-lilly:
I see your point, but I do not know how to solve this one. We would need a whole new parallel graph traversal algorithm, and then some of the most sensitive internals would need refactoring. It is a huge undertaking.

But it may be possible. Somehow, Make already knows how to do this, which means `"Makefile"` parallelism in `drake` should not have this problem. I submitted a [Stack Overflow post](https://stackoverflow.com/questions/47457840/make-with-the-j-option-how-does-make-know-when-to-move-ahead-in-the-dependen) to ask how exactly that works. In the meantime, if you can suggest a parallel graph traversal algorithm, that would help things along.

wlandau-lilly:
But in the short term, your original suggestion of searching the graph further will probably work. It is hard to dispense with parallel stages entirely, but at least we can avoid wasting part of a stage.

The change should take place inside [parallel_stage()](https://github.com/wlandau-lilly/drake/blob/master/R/parallel.R#L17). We need to keep removing up-to-date targets from `config$graph_remaining_targets`  until all our candidate targets are outdated. I am not yet sure how to do this efficiently, but it seems **MUCH** more doable than nuking the whole graph traversal algorithm.

And maybe the algorithm cannot/should not be changed. `Drake` relies on `lapply()`-like functions in most of its backends, which means it uses persistent workers, which means it cannot simply fork new processes on the fly like [Make](www.gnu.org/software/make/). And maybe that's okay since [Make](www.gnu.org/software/make/) is one of the backends anyway (`make(parallelism = "Makefile")`).

wlandau-lilly:
[...]

PR #438, opened 2018-06-29
add future::nbrOfWorkers to future.R

kendonB:
# Summary

future parallelism incorrectly uses the `jobs` argument to `make`. This PR fixes it so that it uses `future::nbrOfWorkers`.

# Related GitHub issues

- Ref: # None.

# Checklist

- [X] I have read the [guidelines for contributing](https://github.com/ropensci/drake/blob/master/CONTRIBUTING.md).
- [X] I have listed any substantial changes in the [development news](https://github.com/ropensci/drake/blob/master/NEWS.md).
- [NA] I have added [`testthat`](https://github.com/r-lib/testthat) unit tests to [`tests/testthat`](https://github.com/ropensci/drake/tree/master/tests/testthat) to confirm that any new features or functionality work correctly.
- [Travis did] I have tested this pull request locally with `devtools::check()`

### DIFF
--- a/R/future.R
+++ b/R/future.R
@@ -169,9 +169,9 @@ running_targets <- function(workers, config){
     unlist
 }
 
-initialize_workers <- function(config){
+initialize_workers <- function(){
   out <- list()
-  for (i in seq_len(config$jobs))
+  for (i in seq_len(future::nbrOfWorkers()))
     out[[i]] <- empty_worker(target = NA)
   out
 }
--- a/R/future.R
+++ b/R/future.R
@@ -1,6 +1,6 @@
 run_future <- function(config){
   queue <- new_priority_queue(config = config)
-  workers <- initialize_workers(config)
+  workers <- initialize_workers()
   # While any targets are queued or running...
   while (work_remains(queue = queue, workers = workers, config = config)){
     for (id in seq_along(workers)){


wlandau:
Do `future`-based `make()`s really ignore `config$jobs`? Now that `drake` uses custom schedulers, it should no longer rely on `future::nbrOfWorkers`. In `make(parallelism = "future")`, `drake` sends out individual futures and manages them with the calling R session. In `make(parallelism = "future_lapply")`, `drake` calls `future_lapply()` to spin up a fixed number of persistent workers in advance, and governing master process sends instructions to those workers. These changes were recent.

kendonB:
My understanding was that you wanted `config$jobs` to just be used for the light parallelization stages, and then defer to the future plan for the number of workers when using `future_lapply`. If so, you want to ignore config$jobs when using future-based makes? 

When I tested _both_ `future` and `future_lapply` using the most recent CRAN version, they both incorrectly used jobs rather than nbrOfWorkers when submitting SLURM jobs.

I guess I'm a bit confused by the first part of your question.

wlandau:
> My understanding was that you wanted config$jobs to just be used for the light parallelization stages, and then defer to the future plan for the number of workers when using future_lapply. If so, you want to ignore config$jobs when using future-based makes?

I have been trying to move `drake` away from staged parallelism as much as possible because we lose parallel efficiency in situations similar to #168. (Can't totally remove it because of situations like https://github.com/ropensci/drake/issues/369#issuecomment-386814320.) As I explained in #437, the reliance on `config$jobs` is part of `drake`'s new behavior for the `future`-based backends.

A lot has changed, and you might find [this recent rOpenSci tech note](https://ropensci.org/technotes/2018/05/18/drake-hpc/) to be a useful high-level summary.

Issue #473, opened 2018-07-20
A new interface for triggers


wlandau:
## The problem

The current trigger interface is limiting. The choices "any", "always", "depends", "command", and "file" miss important use cases. For example, what if we want to check some remote data before deciding whether to build a target? For most projects, we would need to create a special upstream target with the fingerprint of the data and then have the real target depend on that fingerprint.

## Proposal

It would be better to let users write arbitrary R code for each trigger.

```r
drake_plan(
  target1 = {
    get_data(file_in("data.csv"))
    trigger(
      command = FALSE,
      logical = today() == "Tuesday"
    )
  }
  target2 = trigger(custom = remote_hash()),
  target3 = {
    dplyr::bind_rows(target1, tail(target2))
    trigger(
      file = FALSE,
      condition = is_small_enough(target1)
    )
  },
  target4 = {
    summary(target3)
    trigger(
      depends = TRUE,
      change = head(target3, 1)
    )
  }
)
```

Proposed `trigger()` function:

```r
#' @title Set the trigger of a target.
#' @description Use this function inside a target's command
#'   in your [drake_plan()]. The target will rebuild if and only if:
#'   - Any of `command`, `depends`, or `file` is `TRUE`, or
#'   - `condition` evaluates to `TRUE`, or
#'   - `change` evaluates to a value different from last time.
#' @export
#' @seealso [drake_plan()], [make()]
#' @return a list of trigger specification details that
#'   `drake` processes internally when it comes time to decide
#'   whether to build the target.
#' @param command logical, whether to rebuild the target if the
#'   [drake_plan()] command changes.
#' @param depends logical, whether to rebuild if a
#'   non-file dependency changes.
#' @param file logical, whether to rebuild the target
#'   if a [file_in()]/[file_out()]/[knitr_in()] file changes.
#' @param condition R code (expression or language object)
#'   that returns a logical. The target will rebuild
#'   if the code evaluates to `TRUE`.
#' @param change R code (expression or language object)
#'  that returns any value. The target will rebuild
#'   if that value is different from last time
#'   or not already cached.
trigger <- function(
  command = TRUE,
  depends = TRUE,
  file = TRUE,
  condition = NULL,
  change = NULL
){
  stopifnot(is.logical(command))
  stopifnot(is.logical(depends))
  stopifnot(is.logical(file))
  list(
    command = command,
    depends = depends,
    file = file,
    condition = rlang::enexpr(condition),
    change = rlang::enexpr(change)
  )
}
```

[...]

PR #478, opened 2018-07-23
An improved interface for triggers

wlandau:
# Summary

There will be a forthcoming chapter in the manual with the full details. Rather than character strings "any", "always", "command", etc., triggers are now declared with calls to `trigger()`. This allows for much more customization. For example, we can set a target to build only on Tuesdays

```r
drake_plan(
  x = target(
    do_some_stuff(),
    trigger(
      condition = identical(today(), "Tuesday"),
      command = FALSE,
      depend = FALSE,
      file = FALSE
    )
  )
)
```

Or we can make our target sensitive to changes to a remote data source on top of everything else.

```r
drake_plan(
  x = target(
    get_online_data(),
    trigger(change = online_checksum())
  )
)
```

Dependencies are checked in the any trigger code just like in the command. The downside is that the `depend` trigger reacts to changes in functions like `online_checksum()` even if they are not part of the command or its constituent functions.

# Related GitHub issues

- Ref: #473

# Checklist


wlandau:
Tests on real projects pass too.

PR #480, opened 2018-07-24
Make triggers less susceptible to false positives

wlandau:
# Summary

cc @AlexAxthelm. In #478, we decided to detect dependencies from triggers: for example,

```r
f <- function(){
  1 + 1
}
g <- function(){
  TRUE
}
plan <- drake_plan(
  x = target(
    command = f(),
    trigger = trigger(condition = g())
  )
)
vis_drake_graph(drake_config(plan))
```

graph

`g()` may not be part of the command, but we should at least check and process it before we attempt `x`. The problem with this approach is that it is oversensitive to changes to `g()`. Clearly, if we change `g()` to `function(){FALSE}`, `x` should not build. But under #478, it does build because the `depend` trigger is still activated. And we don't want to use `trigger(depend = FALSE)` because we still want to react to changes to `f()`.

The solution in this PR is to ignore changes to dependencies that only show up in the triggers. Implementation details:

- Store a new `"ignore_changes"` `igraph` attribute in `build_drake_graph()` to keep track of these trigger-only dependencies.
- In `dependency_hash()` and `file_dependency_hash()`, exclude everything listed in `"ignore_changes"`.

The implementation admittedly added a little complexity to the code base, and it was not my most elegant work. However, it is consistent with the existing internals and design patterns.

- Ref: #473, #478

# Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,7 +2,7 @@
 
 Please explain the context and purpose of your contribution and list the changes you made to the code base or documentation.
 
-# Related GitHub issues
+# Related GitHub issues and pull requests
 
 - Ref: #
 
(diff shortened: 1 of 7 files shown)


AlexAxthelm:
I like it.

PR #556, opened 2018-10-21
Add a blitz mode

wlandau:
# Summary

In [this article](http://inf.ufrgs.br/gppd/wsppd/2018/papers/proceedings/WSPPD_2018_paper_10.pdf), @guilhermealles and @schnorr attempted to use `drake` to orchestrate parallel tasks in [StarVZ](https://github.com/schnorr/starvz). They found that `drake`'s caching and parallel processing suffered from too much overhead to be useful. In-memory caches and [`clustermq`](https://github.com/mschubert/clustermq)-based parallelism do reduce overhead somewhat.

```r
options(clustermq.scheduler = "multicore") # Or a computing cluster
cache <- storr::storr_environment() # In-memory cache
make(plan, cache = cache, parallelism = "clustermq", jobs = c(imports = 1, targets = 8)
```

However, even this may be inadequate because of all the checking, serialization, hashing, and caching `drake` does in order to skip up-to-date targets.

To eliminate this overhead, `drake` now has a blitz mode. In `make(parallelism = "blitz")`, `drake` just builds all the targets without any checking, hashing, or caching (except for memoized preprocessing). [Benchmarks](https://github.com/ropensci/drake/issues/498#issuecomment-431667017) show a huge improvement in runtime.

In blitz mode, users are responsible for manually saving their results to files, skipping up-to-date work if desired, and trying not to consume too much memory (maybe saving results to files rather than returning large values as targets). So this is really for people who just want `drake` to do parallel computing on a DAG.

Targets build locally in sequence when `jobs = 1` and with [`clustermq`](https://github.com/mschubert/clustermq) when `jobs > 1`. To parallelize locally instead of on a cluster, use `options(clustermq.scheduler = "multicore")`.

It still takes some overhead to turn a `drake` plan into a DAG, but this preprocessing is memoized now. I may also move this functionality to the [`workers`](https://github.com/wlandau/workers) package for cases where we have pre-computed DAGs.

FYI: @kendonB, @krlmlr

- Ref: #498

# Checklist


wlandau:
Update: this is now called "hasty" mode: `make(parallelism = "hasty")`.

Issue #562, opened 2018-10-27
A base R replacement for formatR::tidy_source()


wlandau:
# Background

I am trying to trim down [`drake`'s package dependencies](https://github.com/ropensci/drake/blob/master/DESCRIPTION#L62-L76). I removed 7 just yesterday, bringing the count down from 22 to 15. I am not too concerned about base packages like `utils`, packages in [`r-lib`](https://github.com/r-lib) like [`withr`](github.com/r-lib/withr), or [tidyverse](https://github.com/tidyverse) packages like [`dplyr`](https://github.com/tidyverse/dplyr). And it seems infeasible to remove [`storr`](https://github.com/richfitz/storr), [`igraph`](github.com/igraph/rigraph), or `codetools`. So as I continue on, I hope to find a base R replacement for `formatR::tidy_source()`.

`formatR::tidy_source()` provides a means of [standardizing a target's commands](https://github.com/ropensci/drake/blob/4a85199412d5cea7620f74b43d9356ec3e4f5f6c/R/commands.R#L66-L82). This standardization removes comments, strange indentation, etc. before `drake` decides if a command changed since last `make()`. Clearly, if all we do is add spaces or comments, we don't want `drake` to rebuild the target.

# Requirements

A base R replacement should

1. Style the text's indentation and spacing in a consistent manner.
2. Remove comments.
3. Turn assigment `=`s and `<-`'s into `->`.

# Challenges

R's default parser will take care of (1) and (2), but (3) is tricky.

``` r
library(magrittr)
  parse(text = "z = {f('#') # comment
      
x <- 5
    }",
    keep.source = FALSE
  )[[1]] %>%
    deparse() %>%
    cat(sep = "\n")
#> z = {
#>     f("#")
#>     x <- 5
#> }
```

<sup>Created on 2018-10-28 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>
<sup>Created on 2018-10-28 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

# Rollout

This change will invalidate all targets in all workflows, so I am postponing the release until `drake` 7.0.0 (first half of 2019). We should also add a warning in [`assert_compatible_cache()`](https://github.com/ropensci/drake/blob/4067a755b26fdc055d1ededa13953a2e5e196041/R/cache.R#L513-L542).

wlandau:
To elaborate: this is `drake`'s current behavior.

``` r
drake:::standardize_command
[...]

Issue #345, opened 2018-03-28
R6 object keeps rebuilding


bart1:
First of all thanks for all the good work and quick bug fixes! For some simulations I work with R6 classes, it is a convenient way to keep track of a lot of things in individual based simulations. I encountered issues where targets are rebuild while nothing changed. I tried to make a small reproducible example (here target fenv2 is also rebuild on the second call to make):

```r
require(R6)                                                                                         
FoodEnvironment <- R6Class(                                                                         
   "FoodEnvironment",                                                                                
  private = list(foodItemList = list()),                                                            
  public = list(                                                                                    
    initialize = function(foodDensity,                                                              
                          arenaExtent                                                               
    ) {                                                                       
      private$foodItemList = replicate(                                                           
        round((arenaExtent * 2) ^ 2 * foodDensity),                                               
        runif(2)*arenaExtent*2-arenaExtent, simplify=FALSE                                        
      )                                                                                           
    }
  )                                                                                                 
)                                                                                                   
require(drake)                                                                                      
tmpf<-                                                                                              
  function(foodDensity, arenaExtent) replicate(                                               
    round((arenaExtent * 2) ^ 2 * foodDensity),                                               
    runif(2)*arenaExtent*2-arenaExtent, simplify=FALSE                                        
  )                                                                                           
dd<-drake_plan(                                                                                     
  fenv2=FoodEnvironment$new(.1, 40),                                                   
  fenv=tmpf(.1, 40)                                                                    
)                                                                                    
cache<-recover_cache('asfiiiiif')                                                                      
drake::make(dd, cache=cache, verbose=4)                                                             
drake::make(dd, cache=cache, verbose=4)  
```

This gives the following results:
```r
> source('~/.active-rstudio-document')
Loading required package: R6
Loading required package: drake
connect 4 imports: FoodEnvironment, tmpf, cache, dd
connect 2 targets: fenv2, fenv
check 4 items: FoodEnvironment, replicate, round, runif
import FoodEnvironment
import replicate
import round
import runif
check 1 item: tmpf
import tmpf
check 2 items: fenv, fenv2
target fenv
target fenv2
Unloading targets from environment:
  fenv2
  fenv
connect 4 imports: FoodEnvironment, tmpf, cache, dd
connect 2 targets: fenv2, fenv
check 4 items: FoodEnvironment, replicate, round, runif
import FoodEnvironment
import replicate
import round
import runif
check 1 item: tmpf
import tmpf
[...]

PR #563, opened 2018-10-28
A base R replacement to formatR::tidy_source()

wlandau:
# Summary

**NOTE: this PR will invalidate all targets in all workflows, so I am waiting until the next major release (version 7.0.0) to merge this it!**

In the interest of reducing `drake`'s dependencies on other packages, this PR introduces a replacement to `formatR::tidy_source()` that only depends on packages `base` and `utils`. Thanks to @hrbrmstr and @lorenzwalthert for their solutions in #562.

- Ref: #562

# Checklist


lorenzwalthert:
It's a detail, but since the output of the function has changed, maybe a renaming would be appropriate to reflect the fact that it computes a hash?

wlandau:
It's a good point. However, now that you mention it, I am not totally settled on the output format. `drake` serializes and hashes (via `storr`) whatever gets returned from `standardize_command()` (which calls `standardize_code()`). I am not sure if it is faster to

1. Hash once and then hash and serialize the first hash (current behavior) or
2. Return the table, which gets serialized and hashed once.

In (2) we hash only once, but the serialization is applied to a larger object of varying size. We have a long time to decide. I plan to keep this PR open at least until we are sure the next release will be version 7.0.0.

lorenzwalthert:
Good point. Maybe you are right and we should not hash twice if possible. Can you quantify the speed benefits of (2) vs (1)?

wlandau:
A good thing to check. The times below look about the same between options. I think option 1 is a good choice for now because it should take up less memory.

``` r
library(magrittr)
library(microbenchmark)

# Return tokenization info for code string x
standardize <- function(x){
  if (!length(x)){
    return(as.character(NA))
  }
  x <- deparse(parse(text = as.character(x), keep.source = FALSE)[[1]]) %>%
    paste(collapse = "\n")
  info <- parse(text = x, keep.source = TRUE) %>%
    utils::getParseData(includeText = TRUE)
  change <- info$token == "LEFT_ASSIGN"
  info$text[change] <- "="
  info$token[change] <- "EQ_ASSIGN"
  info[info$token != "COMMENT" & info$terminal, c("token", "text")]
}

# Try option 1: hash, serialize the hash, and then hash again.
option_1 <- function(x, key, cache){
  standardize(x) %>%
    lapply(FUN = paste, collapse = " ") %>%
    paste(collapse = " >> ") %>%
    digest::digest(algo = "sha256", serialize = FALSE) %>% # First hash
    cache$set(key = key) # Serialize and hash via `storr`
}

# Try option 2: let storr do all the serialization and hashing.
option_2 <- function(x, key, cache){
  standardize(x) %>%
    c() %>%
    cache$set(key = key) # serialization and hashing.
}

# Borrow and modify the body of standardize() as example code
# Create text of variable sizes, up to 32 times the original.
random_code <- function(){
  text <- deparse(body(standardize))
  n <- sample.int(n = 5, size = 1)
  for (i in seq_len(n)){
    text <- c(text, text)
  }
  text <- c("{", text, "}")
  paste(text, collapse = "\n") %>%
    gsub(pattern = "ASSIGN", replacement = as.character(runif(1)))
}

# One benchmark
benchmark <- function(fun){
  set.seed(0)
  times <- 100
  keys <- as.character(seq_len(times))
  code <- replicate(times, random_code())
  set.seed(0)
  cache <- storr::storr_rds(tempfile())
  on.exit(cache$destroy())
  loop <- function(){
    for (i in seq_len(times)){
      fun(x = code[i], key = keys[i], cache = cache)
    }
  }
  microbenchmark(loop(), times = 100)
}

# Run the benchmarks
print(benchmark(option_1))
#> Unit: seconds
#>    expr      min      lq     mean   median       uq     max neval
#>  loop() 6.708124 6.89136 7.010094 6.949077 7.093694 7.58899   100
print(benchmark(option_2))
#> Unit: seconds
#>    expr      min       lq     mean   median       uq      max neval
#>  loop() 6.813503 6.916124 7.042026 6.998259 7.084101 7.605348   100
```

Created on 2018-10-30 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)

lorenzwalthert:
Not sure if related but I remember when I removed a lot of old big targets from the cache that I did not use anymore in my workflow plan, drake got considerably faster.

wlandau:
Hmm... I would be surprised if the size of targets mattered, but the number sure does. I bet there are still many linear-time lookups in `drake`'s code base.

wlandau:
Edit: size could also matter in the sense that immediate dependencies need to be loaded into memory. If there is more slowdown with respect to the size of targets than that can explain, something is wrong.

lorenzwalthert:
May I ask what the rationale behind replacing `%>%` in 69ff455 is? Is drake moving away from `%>%`?

wlandau:
Yes: #570, #571. Profiling the code with `provfis` is much easier this way. No more tall stacks of `fseq()` and `freduce()`. I am already learning a lot now that that clutter is gone.

wlandau:
Example: #572. Would not have been obvious if the call stack of `%>%` had interfered.

lorenzwalthert:
Ok, thanks for the explanation. However, as this seems quite a fundamental problem, so maybe it's worth filing an issue in rstudio/profvis to disentangle? Other option: use stlyer to remove pipe and then use the generated code for profiling. Quite ambitions probably. Also, I remember @krlmlr doing profiling for styler in the tibble dependency using https://github.com/r-prof/jointprof or a similar tool he has been working on with an R Consortium grant where he was able to profile recursions. Potentially the feature of *unpiping* could be implemented in this profiler?

wlandau:
Good idea. See https://github.com/rstudio/profvis/issues/102.

lorenzwalthert:
👍

krlmlr:
I guess we have to wait for a better _magrittr_ for nicer call stacks when using the pipe. There's code in the dev version of _magrittr_ that would help un-piping, but AFAIK it's not on CRAN yet.

wlandau:
- [ ] Note to self: just before merging, set the minimum [compatible cache version](https://github.com/ropensci/drake/blob/33328cc5824fcbca2f9d676a01af4e6b1dbe44e2/R/cache.R#L528) to the release just prior to 7.0.0.

lorenzwalthert:
On a related note: I don't know how this is handled now but when i updated drake the last time, I had to re-build the whole cache. I assume this won't be any different this time? I think a warning should be issued after the installation so people could still go back to the old installation for any reason and use the built cache. Just in case it is not possible to re-build the cache for some reason, or it needs so much time that the user may want to defer it. If this is a thing, I think we best open a new issue.

wlandau:
> On a related note: I don't know how this is handled now but when i updated drake the last time, I had to re-build the whole cache. I assume this won't be any different this time?

If we go through with #562 and #563, yes. I am still wondering if the change is worth the removal of a single dependency, and I could be convinced either way. If we are going to do it, it is better sooner rather than later, especially since 7.0.0 will have large changes already. But it is also important to fulfill `drake`'s promise of keeping the right targets up to date.

> I think a warning should be issued after the installation so people could still go back to the old installation for any reason and use the built cache.

Seems like we would want to control the installation process: look at the user's prior version of `drake`, install the package, and give a warning if the old and new versions span a change in the cache format. Is that possible to do in a package? It's a good idea, and I have often wanted to add special installation commands for different reasons. Either way, I think it is important to look at individual projects too.

wlandau:
Looking back at this issue, I have a couple comments:

1. I think now is the time to move away from `formatR`. Because of #630, #633, and #641 (which are well worth the speed improvements) the next release will invalidate targets in old projects anyway.
2. I no longer think we should coerce `EQ_ASSIGN` to `LEFT_ASSIGN`. The reason is consistency: functions already get fingerprinted by deparsing, which does not change `=` to `<-`. The behavior for commands in the plan should agree with the behavior for functions. So I think we should go with a simple `deparse(parse(text = as.character(command), keep.source = FALSE)[[1]])` instead of `utils::getParseData()`.

wlandau:
A positive note on speed:

``` r
text <- paste(deparse(digest::digest), collapse = "\n")
microbenchmark::microbenchmark(
  deparse(parse(text = text, keep.source = FALSE)[[1]]),
  formatR::tidy_source(
    source = NULL,
    comment = FALSE,
    blank = FALSE,
    arrow = TRUE,
    brace.newline = FALSE,
    indent = 4,
    output = FALSE,
    text = text,
    width.cutoff = 119
  )
)
#> Unit: microseconds
#>                                                                                                                                                                             expr
#>                                                                                                                            deparse(parse(text = text, keep.source = FALSE)[[1]])
#>  formatR::tidy_source(source = NULL, comment = FALSE, blank = FALSE,      arrow = TRUE, brace.newline = FALSE, indent = 4, output = FALSE,      text = text, width.cutoff = 119)
#>       min        lq      mean    median       uq      max neval
#>   551.514  566.5695  589.0592  572.3665  594.850  749.532   100
#>  3728.347 3897.0460 4218.4730 3996.0425 4161.814 7441.496   100
```

<sup>Created on 2019-01-01 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

wlandau:
Hmm... `deparse()` is slow. In 8ab9d13df64b228b5ba859d8af19e754337757a4, I decided to just removed all the attributes (including `srcref`, `srcfile`, and `wholeSrcref`) and return the language object without deparsing it. The speedup was about 2-3% for [this workflow](https://github.com/wlandau/drake-examples/tree/master/overhead). @lorenzwalthert, is there anything wrong with this approach? Any hidden pitfalls I am missing?

wlandau:
That one deparse() is for imported functions because hashes of functions are brittle: https://github.com/ropensci/drake/issues/345#issuecomment-385073073. My question: are the hashes of expressions, symbols, and language objects in general just as brittle? If x is a language object, there anything that would make digest(x) unreliable. We can assume attributes(x) is NULL. Also cc @wch.

wlandau:
I actually found the answer to my question while refining #233 today. Strange things happen with language objects in R version 3.4.4. The two language objects below *should* be the same, but they have different internals because they were generated in different ways. `drake` now uses an internal `safe_deparse()` function to standardize commands (demonstrated below). The change invalidates targets again, but I think it is the right call, and I am glad I caught it before the next CRAN release.

``` r
x <- rlang::syms(c("a", "b"))
names(x) <- c("a", "b")

lang1 <- substitute(f(x), list(x = x))
lang2 <- quote(f(list(a = a, b = b)))

print(lang1)
#> f(list(a = a, b = b))
print(lang2)
#> f(list(a = a, b = b))

library(digest)
digest(lang1)
#> [1] "869aef282c5349417ab12590f1359de5"
digest(lang2)
#> [1] "b6630dff6ebb37dac41c6164381efcc7"

deparse(lang1)
#> [1] "f(structure(list(a = a, b = b), .Names = c(\"a\", \"b\")))"
deparse(lang2)
#> [1] "f(list(a = a, b = b))"

print(drake:::safe_deparse)
#> function (x) 
#> {
#>     paste(deparse(x, control = c("keepInteger", "keepNA")), collapse = "\n")
#> }
#> <environment: namespace:drake>

drake:::safe_deparse(lang1)
#> [1] "f(list(a = a, b = b))"
drake:::safe_deparse(lang2)
#> [1] "f(list(a = a, b = b))"

digest(drake:::safe_deparse(lang1))
#> [1] "c19a8f963e567ebb171494483dfb1697"
digest(drake:::safe_deparse(lang2))
#> [1] "c19a8f963e567ebb171494483dfb1697"
```

<sup>Created on 2019-01-21 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

wlandau:
Hmm... the same example on R 3.5.2 gives 2 different answers.

``` r
x <- rlang::syms(c("a", "b"))
names(x) <- c("a", "b")

lang1 <- substitute(f(x), list(x = x))
lang2 <- quote(f(list(a = a, b = b)))

deparse(lang1, control = NULL)
#> [1] "f(list(a, b))"
deparse(lang2, control = NULL)
#> [1] "f(list(a = a, b = b))"
```

<sup>Created on 2019-01-21 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

Hopefully deparsing prevents most other edge cases. Other remarks:

1. For #233, `combine()` should not try to produce named lists. In [this example](https://ropenscilabs.github.io/drake-manual/plans.html#combine), the command for `larger` should always stay `do.call(rbind, list(data_1_3, data_2_4))`. Outputting `do.call(rbind, list(data_1_3 = data_1_3, data_2_4 = data_2_4))` might seem more useful, but the names might get dropped for old versions of R. (Either that or you get an unpredictable `structure(..., .Names = ...)`.
2. At one point, @krlmlr suggested that `drake` might be better off using language objects for commands instead of character strings. But until we figure out a better way to handle language objects, I am actually kind of glad my naivete accidentally made `drake` a little more robust.

Issue #575, opened 2018-11-07
Bare-bones scheduling and caching


krlmlr:
For some use cases of _drake_, I'd rather just supply the dependency graph, the commands (as quosures or perhaps [crates](https://cran.r-project.org/web/packages/carrier/index.html)), and a list of targets to build. I would expect _drake_ to take care of

- caching and retrieving objects, making them available for the expression/function
- topological sorting, scheduling and running the code
- live updates of the dependency graph and of the targets to be built (later, for #304)

but really not more. This will remove all overhead from analyzing the expressions, leaving it to a higher level in the API and perhaps other frontend packages.

I believe that separation along these lines may help the overall architecture, and eventually lead to a clean separation into the _workers_ package -- as a first step a lower-level API (perhaps with a distinct prefix) would do the job.

I've seen hasty parallelism, my proposal feels like hasty plus object storage minus parsing. Does that make sense?

wlandau:
I like the idea. Do you think we can pilot it from within hasty mode first? Object storage is just a matter of supplying the right `hasty_build` function to `make()` (see `default_hasty_build()`).

And I like your plan for `workers`. I was planning to move hasty mode there eventually, and I think it is worth seeing how bare-bones caching/scheduling plays out in `drake` itself.

As pre-work for live updates to the dependency graph, I am reopening #440. Supplying a graph and cleaning up the architecture will be much simpler if the graph does not need special attributes. (The current behavior is to generate an `igraph` of jobs and attach more verbose dependency information to the nodes.) My current thinking is to have the following major data structures:

- The workflow plan data frame with dependencies *implicitly* specified.
- The lookup structure from #440 with the dependencies *explicitly* specified as in a `Makefile` (and listed by type: targets, files, dependencies from triggers, etc.) I am thinking this lookup structure (name pending) should be the direct outcome of the code analysis. Each element should be like one of the objects returned from `deps_target()`.
- A simple `igraph` of jobs.
- A priority queue.

My major development goals for `drake` right now are to reduce overhead and simplify the architecture, both for bare-bones execution and the rest of the package.

wlandau:
> Are you suggesting to expose an API to manually construct the lookup structure? 

[...]

Issue #504, opened 2018-08-16
Calculate inputs once during plan construction?


gmbecker:
@wlandau it looked from your talk that the plan is a data.frame that has the target-code linking but it doesn't have the the inputs for the expression. Those aren't going to change for as long as the expression doesn't change, so why not calculate it up front and put it in the plan df right away. You can use a list column for this to get around the fact that an expression can have 0-many inputs.  Apologies if I missed something under the hood that avoids this repetition already under the hood.

wlandau:
Glad we got a chance to talk about this in person yesterday. I got pretty far along in development before I realized code analysis was a bottleneck, and `make()` indeed repeats the dependency detection step in most cases. In the current `drake`, there are a couple existing ways to speed things along:

1. Pre-calculate a `drake_config()` object and pass it to the `config` argument of `make()`, or
2. Pre-calculate the dependency graph with `build_drake_graph()` and pass it to the `graph` argument of `drake_config()` or `make()`. This graph is an `igraph` of targets/jobs with specific dependency information stored in vertex attributes.

These approaches avoid the redundant computation of not only the code analysis, but also the construction of the dependency/job `igraph` afterwards, both of which can be bottlenecks for jobs with lots of targets. However, if the user changes an imported function enough to change the dependency relationships, they will need a new `drake_config()` or `igraph` object all over again. With your approach, users would still be free to change custom functions, and we could still save time during the initial stages of `make()`.

So far, I see a couple ways to do this, and there may be more.

#### `finalize_plan()`

We could implement your proposed `finalize_plan()` function to add each target's dependency information as a list column in the plan. Users would construct the usual plan with `drake_plan()` or some custom method, call `finalize_plan()`, and then call `make()`. That way, users would be free to change their custom (non-command) functions and still save time on the preprocessing. While this does seem to strike a great balance between time savings and flexibility for long plans, I am not sure about adding another required step in the workflow for everyone.

#### A new `deps` argument of `make()`

We could optionally pass the list of target dependencies into a new `deps` argument of `make()`/`drake_config()`. In my current thinking, `deps` would behave just like the `graph` and `config` arguments. The default behavior of `make()` would not change, but power users could supply a `deps` list with all the pre-calculated dependencies of the targets. We could have a new `deps_plan()` function to create this dependency list and group it alongside `deps_code()` and `deps_targets()` in the documentation. 

[...]

Issue #440, opened 2018-06-29
A new internal fast-lookup data structure for target information


wlandau:
### Lessons from #435 and #283:

- `drake` spends a lot of time repeatedly looking up target-level information it should already know.
- The dependency network is inflexible.

### Explanation

I think these and similar problems come from the limitations of `drake`'s data structures. The `plan` is great for users, but lookup is slow, and it contains only indirect information about each target's dependencies. The `graph` has complete dependency information, but it does not have other metadata (`igraph` attributes are a pain), and traversing the whole graph for dependencies is slow.

### Solution

I want to put all target-level metadata in a unified fast-lookup data structure. This new `config$meta` will replace individual outputs of `drake_meta()` flying around, and it will decrease reliance on the `plan` and `graph`. For [fast lookup](https://stackoverflow.com/questions/34232865/r-fast-single-item-lookup-from-list-vs-data-table-vs-hash), I want to base this data structure on an environment. For convenience, I want it to have a `storr`-like API. Best of both worlds: a [`storr_environment()` cache](http://richfitz.github.io/storr/reference/storr_environment.html). (Thank you @richfitz for making it available!)

There is a lot of refactoring to do, but I think this will make `drake` a whole lot faster and give it a cleaner code base.

wlandau:
For the sake of #283, we should probably distinguish between the build schedule and the constituent files and objects. Maybe we need separate `storr_environment()`s:

1. `config$meta`: a `storr_environment()` for individual data objects. Should have the following `storr` namespaces:
    - `name`: character string
    - `imported`: logical, imported from the user's workspace/environment or generated with a `drake` command?
    - `file`: whether the item is a file or simply a generic R object.
    - `missing`: whether the item is present in `config$envir`
    - `hash`: the object's fingerprint.
    - `mtime`: file modification time (`NULL` for non-files)
2. `config$protocol`:
    - `"name"`: name of the step in the schedule, usually a target or import
    - `"command"`: command from the `drake_plan()` (`NULL` for imports)
    - `"seed"`: RNG seed
    - `"trigger"`: the trigger from the `drake_plan()` (or the default trigger)
    - `"object_in"`: names of non-file dependencies
    - `"file_out"`: names of any file outputs (potentially more than one)
    - `"file_in"`: names of any non-`knitr` input files
    - `"knitr_in"`: names of any `knitr` input files
    - `"dependency_hash"`: A hash used to trip the `depends` trigger.
[...]

PR #576, opened 2018-11-09
Intermediate data structure to store all the code analysis

wlandau:
# Summary

One major difference between `drake` and most other pipeline tools is that users declare dependencies *implicitly* in their plans and workspaces. Other pipeline tools like Make declare dependencies *explicitly* with rigorous build rules. The jump from *implicit* components to *explicit* dependency information is time-consuming and nontrivial, so in addition to the workflow plan data frame, I think we need a special data structure whose sole purpose is to store the full dependency detection results of the code analysis. I believe this will pave the way to make `drake` cleaner and faster in the long run.

# Implementation

```r
load_mtcars_example()
config <- drake_config(my_plan)
str(config$ordinances)
```

This data structure is a list, and each element corresponds to a target or import. Each target's corresponding element of `ordinances` has the complete specification of immediate dependencies (other targets, output files, input files, etc.) the trigger, and all the fields of the workflow plan data frame for fast lookup. 

`config$ordinances` is a list. I thought about making it an environment so lookup would be even faster, but then we would not be able to use `lightly_paralleize()` to run through it. I am open to other suggestions. Either way, lookup should be constant-time, and we might see a little improved performance.

# Plans for merging

I will be on vacation from November 15 through 27, and I will wait to merge this PR until I come back. I want to be present to address any bug reports, and I want more time to think about the name. "Ordinances" is a good initial name because it is distinctive and thus easy to `sed` out later, but I am not totally sold on it.

# Related discussion and issues

This PR implements #440, and I believe this is essentially what @gmbecker originally suggested in #504. #498 and #575 is also related.

cc @krlmlr 

# Checklist


wlandau:
A quick performance study using the [overhead example](https://github.com/wlandau/drake-examples/tree/master/overhead) (ref: #498):

```r
library(drake)
library(microbenchmark)
library(storr)

# See vis_drake_graph(drake_config(plan(8)), targets_only = TRUE)

plan <- function(n){
  plan <- drake_plan(target_1 = 1)
  for (i in seq_len(n - 1) + 1){
    target <- paste0("target_", i)
    dependencies <- paste0("target_", seq_len(i - 1))
    command <- paste0("max(", paste0(dependencies, collapse = ", "), ")")
    plan <- rbind(plan, data.frame(target = target, command = command))
  }
  plan
}

cache <- function(n){
  path <- file.path(tempdir(), paste0("cache", n))
  cache <- storr_rds(path, mangle_key = TRUE)
  clean(cache = cache)
  cache
}

overhead <- function(n){
  make(plan(n), cache = cache(n), verbose = 4)
}

system.time(overhead(1024))
```

Runtime with https://github.com/ropensci/drake/pull/576/commits/a63f1dc0273a834647b79614805c45034cdf795c (this PR):

```r
   user  system elapsed 
161.540   6.520 171.298
```

Runtime with e356a0d38c43c30164132c2c595615199b12f59d (current master branch):

```r
   user  system elapsed 
160.460   6.360 168.793 
```

So no clear performance gain or loss at the moment. Either this PR does not speed things up or other bottlenecks are masking the boost.

Issue #593, opened 2018-12-03
Remove dependency on dplyr


wlandau:
Of all `drake`'s strict package dependencies, `dplyr` is the heaviest. Challenges:

- Replace `group_by() %>% do()` etc. with base R.
- Substitute `bind_rows(x, y)` with something that can handle different columns between `x` and `y`.

bpbond:
Happy to help here if useful.

wlandau:
Thank you for the offer, @bpbond! I would really appreciate help reducing package dependencies in both "Imports:" and "Suggests:". I knew extricating `dplyr` and `tidyselect` would be messy, so I took care of them myself in #594.

In "Imports:", `purrr` and `R6` are the easiest to tackle. Don't worry about `pkgconfig` or `formatR`, those two will go away on their own in the next major release (see #564 and #563). `codetools` is super difficult, and I am working on replacing it via #573. Other than that, I have good reasons for keeping `digest`, `igraph`, `rlang`, `storr`, `tibble`, and `utils` where they are.

The "Suggests:" list is massive. At this point, `abind`, `MASS`, `DBI`, and `RSQLite` are only used in tests. Likewise, `dplyr` and `ggplot` should only be needed in examples (especially https://github.com/wlandau/drake-examples). I may remove `ggraph` at some point because `drake` can now print static graphs automatically with `webshot` (see the documentation of the `file` argument of `vis_drake_graph()`). `future.apply` and `txtq` are likely to go away via #561 at some point.

bpbond:
I'll take a look a de-`purrr`ing the code.

wlandau:
Thanks, that would really help!

wlandau:
I am so glad you are still interested in helping!

## Profiling

Now that the "Imports:" list is much smaller, I think the number 1 priority is speed. Computational overhead is still a major weakness, and with `magrittr` and `dplyr` gone from the stack, profiling with [`profvis`](https://rstudio.github.io/profvis/) is much easier than in the past. In other words, there has never been a better time to speed up `drake`. You could make a huge impact this way.

The [overhead example](https://github.com/wlandau/drake-examples) (ref: #498) is a toy workflow that isolates and exaggerates `drake`'s busywork, and the files contain a GNU Make counterpart for comparison.

## More cleanup and streamlining

As long as we are thinking about shortening the "Suggests:" list, maybe we could use a base R replacement for `rprojroot::find_root()`. It is only called once in `R/test.R`.

[...]

PR #603, opened 2018-12-07
Remove rprojroot dependency

bpbond:
# Summary

Per suggestion, remove `rprojroot` as a suggested package. This meant implementing a simple version of `rprojroot::find_root` in `unit_test_files()`.

- Ref: #593 

# Checklist

- [X] I have listed any substantial changes in the [development news](https://github.com/ropensci/drake/blob/master/NEWS.md).
- [X] I have added [`testthat`](https://github.com/r-lib/testthat) unit tests to [`tests/testthat`](https://github.com/ropensci/drake/tree/master/tests/testthat) to confirm that any new features or functionality work correctly.
- [X] I have tested this pull request locally with `devtools::check()`
- [X] This pull request is ready for review.
- [X] I think this pull request is ready to merge.

### DIFF
--- a/R/test.R
+++ b/R/test.R
@@ -8,14 +8,14 @@ testrun <- function(config) {
   set_test_backend()
   invisible(
     make(plan = config$plan, targets = config$targets, envir = config$envir,
-      verbose = config$verbose, parallelism = config$parallelism,
-      jobs = config$jobs,
-      packages = config$packages, prework = config$prework,
-      prepend = config$prepend, command = config$command,
-      cache = config$cache, lazy_load = config$lazy_load,
-      session_info = config$session_info,
-      fetch_cache = config$fetch_cache,
-      caching = config$caching
+         verbose = config$verbose, parallelism = config$parallelism,
+         jobs = config$jobs,
+         packages = config$packages, prework = config$prework,
+         prepend = config$prepend, command = config$command,
+         cache = config$cache, lazy_load = config$lazy_load,
+         session_info = config$session_info,
+         fetch_cache = config$fetch_cache,
+         caching = config$caching
     )
   )
 }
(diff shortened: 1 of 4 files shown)


R/test.R
```diff
@@ -94,9 +94,20 @@ set_test_backend <- function() {
...
-  root <- rprojroot::find_root(criterion = "DESCRIPTION", path = path)
-  file.path(root, "tests", "testthat")
+  # find the package root
+  p <- normalizePath(dirname(path))
+  MAX_DEPTH <- 100
+  CRITERION <- "DESCRIPTION"
+  for (i in seq_len(MAX_DEPTH)) {
+    if (length(list.files(p, pattern = CRITERION))) {
+      # found criterion file; make sure it's ours
+      if (grepl("drake", readLines(file.path(p, CRITERION), n = 1))) {
```
wlandau: Could we change the grep pattern to `"^Package: drake$"` and search all the lines of the `DESCRIPTION`?

```r
if (grepl("^Package: drake$", readLines(file.path(p, CRITERION)))) {
```

https://github.com/tiernanmartin/drakepkg and/or https://github.com/lbusett/draketest may go to CRAN at some point, and their `DESCRIPTION` files would also fit the pattern.

tests/testthat/test-testing.R
```diff
@@ -115,3 +112,35 @@ test_with_dir("test_scenarios()", {
...
+  # DESCRIPTION without 'drake' in first line
+  writeLines(
+    text = "Package: drakebutnotdrake",
+    con = "DESCRIPTION"
+  )
+  expect_error(unit_test_files(wd))
+
+  # Shouldn't find anything if there's no DESCRIPTION
+  unlink("DESCRIPTION")
+  expect_error(unit_test_files(wd))
```
wlandau: Last couple things:

1. This test executes in a `tempfile()` subdirectory of `tempdir()`. if there is already a copy of `drake`'s `DESCRIPTION` file somewhere in `tempdir()` or its ancestors, the test will incorrectly fail. Would you add a `max_depth` argument to `unit_test_files()` (with default 100) and call `unit_test_files(wd, max_depth = 1)` here and line 141? Maybe we should also write `unit_test_files(subdir, max_depth = 2)` in line 134.
2. Could we use lowercase letters for `criterion` and `max_depth`?

Issue #619, opened 2018-12-15
Lock the user's enviornment during make()


wlandau:
Ideally, `make(plan)` should not modify the user's environment because that would risk invalidating targets. Relevant: #615. Incomplete/failed solutions:

1. Just recommend `make(session = callr::r_vanilla())`. Seems to work, but it is incompatible with the `parallel` package, so it is not a universal fix.
2. Create a separate environment for execution: #296. Problems:
    - The `environment<-` hack seems to break certain kinds of functions, e.g. from `Vectorize()`.
    - Could use up non-negligible computer memory and time.

I think we can try the following. The checked items are completed in the `619` branch.

- [x] Just before building the targets:
    - [x] Define `config$eval <- new.env(parent = config$envir)`.
    - [x] Lock `config$envir` with `lockEnvironment()`.
- [x] Use `config$eval` for:
    - [x] Executing commands `R/run.R`.
    - [x] Loading/unloading targets in `R/envir.R`.
- [x] At the end of `make()`, unlock `config$envir` again with some C code. Ideally, this should not require `Rcpp`. Options:
    - https://gist.github.com/wch/3280369#file-unlockenvironment-r
    - https://stackoverflow.com/questions/25910778/unlockenvironment-implemented-via-rcpp-instead-of-inline/25922051#25922051.

wlandau:
More to-do items:

- [ ] Hijack `e$message` in [`with_call_stack()`](https://github.com/ropensci/drake/blob/master/R/run.R#L56-L60) to append informative text if the message has "cannot add bindings to a locked environment".
- [ ] Lock the `config$envir` on parallel workers if `parallelism` is `"clustermq"`, `"future"`, or `"hasty"`.

wlandau:
Moving this to 7.0.0. I will have more to say in the PR.

wlandau:
To be clear: `config$eval` is already in place, and the user's environment (`config$envir`) is no longer modified by `drake`. It is only the strict environment locking that I am putting off until version 7.0.0.

wlandau:
This change is super important for reproducibility (and to ease frustration in cases like #615). However, it may disrupt many existing workflows, so I have will also add a `lock_envir` option to `make()` and `drake_config()`. It will be `FALSE` by default initially and then switch to a default of `TRUE` in `drake` version 7.0.0.

wlandau:
- [ ] Last item: set `lock_envir` to `TRUE` in `drake` 7.0.0.

[...]

Issue #615, opened 2018-12-13
Targets get outdated when running make() in multiple R session


pat-s:
Hi Will,

looks like I need your help again.
A weird usage problem that I could not reproduce using the included examples in the package.
Sorry for the particular unspecific wall of text but this thing drives me mad (even though it might be my own fault).
Don't feel the need to reply to every sentence of this post, it's more thought to be a collection of anomalies that I have observed lately and stored in my mind.

Currently I do the following:

1. Run [drake.R](https://github.com/pat-s/pathogen-modeling/blob/master/scripts/drake.R)
2. Run `make(plan, <some target>)`
3. Open a new R session 
4. Do again 1) and 2)

So often multiple targets run in parallel in different R sessions. Both R session operate on the same `drake_plan`.

I expected the `drake_config` to be the same for all since its stored in the cache and retrieved from there?
However, often I get different outdated/up-to-date targets in both R session, not related to the targets which are going to be build in that specific R session.

Also, if one `make()` call finishes, the changes sometimes get picked up by the other session, sometimes not. 
Even if I again init the plan or the config.
In addition, I always get different outcomes for `outdated()` and `outdated(config)`.
The worst case the happens is: Sometimes one session completely starts from scratch and invalidates all already built targets. 
I've tried for days now to find a pattern in this behavior but did not succeed.

Similar: If I start two fresh sessions and have ~ 70% of the targets built. I start both with outdated targets of the plan and sometimes one R session will again start from scratch and invalidate all targets. Even though I did not change anything in the plan or any dependency (this is particular problematic since the whole workflow will take > 50h).

This also holds true for calling `vis_drake_config()`: First I always called it with a previously created `config = drake_config())`. 
Then I realized that the result does not always match my actual config so I switched to `vis_drake_config(drake_config())`. 
But this also showed a wrong behavior sometimes (e.g. false positive outdated targets, false negatives failed targets that have already been rebuilt in a previous call). 

I read the manual about the "config" stuff and other possibly related sections but could not find a solution.

Is it not recommended to run different targets in different R sessions in parallel?
Is there a best practice guide on how to not corrupt the cache and the config? 
Even when just using one R session I faced faulty configs and outdated/up-to-date results or targets have been rebuilt even if there status (as a dependency) was "up-to-date".

PS: If you give it a try with the linked repo, just FYI: It downloads around 4 GB of data for the first targets.

wlandau:
I will do what I can to help.

[...]

PR #621, opened 2018-12-16
Set lock_envir to TRUE by default

wlandau:
cc @krlmlr, @pat-s, @rkrug

## Work up to this point

If a user implements impure commands/functions as in https://github.com/ropensci/drake/issues/615#issuecomment-447585359, the act of building a target threatens to change upstream dependencies. To solve this problem, I implemented a `lock_envir` argument to `make()` and `drake_config()`. If set to `TRUE`, `drake` locks the user's environment and all its non-hidden bindings to prevent imported functions etc. from being modified.

Also, `drake` itself no longer inserts built targets into the user's environment. Instead, it uses a separate `config$eval <- new.env(parent = config$envir)`.

## This PR

I believe `make(lock_envir = TRUE)` is super important for reproducibility. However, it is a disruptive change, and existing impure projects may error out. That is why I think `drake` version 7.0.0 will be a great time to merge this PR to set the `lock_envir` default value to `TRUE`.

# Summary

Please explain the context and purpose of your contribution and list the changes you made to the code base or documentation.

- Ref: #619, #269

# Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,10 @@
+# Version 7.0.0
+
+## Breaking changes
+
+- Set the default value of `lock_envir` to `TRUE` in `make()` and `drake_config()`. So `make()` will automatically quit in error if the act of building a target tries to change upstream dependencies.
+
+
 # Version 6.2.1.9000
 
 ## Bug fixes
--- a/R/config.R
+++ b/R/config.R
@@ -494,7 +494,7 @@ drake_config <- function(
   hasty_build = drake::default_hasty_build,
   memory_strategy = c("speed", "memory", "lookahead"),
   layout = NULL,
-  lock_envir = FALSE
+  lock_envir = TRUE
 ) {
   force(envir)
   unlink(console_log_file)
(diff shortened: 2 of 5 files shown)


wlandau:
## How it works

`lock_environment()` locks an environment an all its non-hidden bindings:

```r
lock_environment <- function(envir) {
  lockEnvironment(envir, bindings = FALSE)
  lock_these <- ls(envir, all.names = FALSE)
  skip_these <- ".Random.seed" # Probably unnecessary, but just in case we set all.names to TRUE above
  lock_these <- setdiff(lock_these, skip_these)
  lapply(X = lock_these, FUN = lockBinding, env = envir)
  invisible()
}
```

`unlock_environment()` uses a bit of C code I borrowed and modified from https://gist.github.com/wch/3280369#file-unlockenvironment-r and https://github.com/SurajGupta/r-source/blob/master/src/main/envir.c.


The executing process locks the environment just before the [target's command is executed](https://github.com/ropensci/drake/blob/e2cb2724252704b7cdd98816cfd7271fa8c53e9f/R/run.R#L78) and unlocks the environment afterwards. In addition, it *blocks* the lock so no other process can come along and lock/unlock it (guarantees the environment stays locked while the command is running, and it does not seem to cause a bottleneck for parallel computing).

https://github.com/ropensci/drake/blob/e2cb2724252704b7cdd98816cfd7271fa8c53e9f/R/run.R#L65-L74

PR #671, opened 2019-01-12
Deprecate (hide) internal functions

### LINKED ISSUE #644: Trim down the API for 7.0.0
`drake`'s API has too many functions. Many of them should be deprecated.

bpbond:
Closes #644 

# Summary

Deprecate and create a second internal "_"-appended function for:
* `analysis_wildcard()`
* `cache_namespaces()`
* `cache_path()`
* `dataset_wildcard()`
* `drake_meta()`
* `process_import()` (not on the original list in #644 )
* `drake_palette()`
* `drake_tip()`
* `in_progress()`
* `recover_cache()`
* `target_namespaces()`

Fully deprecated, i.e. no internal "_" function created:
* `check_plan()` 

- Ref: #644 

# Checklist


wlandau:
Thanks, I will look at this later today.

By the way, if the lint bot seems excessive at times, the lintr package itself might save you some trouble. [This bash function](https://github.com/wlandau/dotfiles/blob/94825dfdbc4243c0ff2bd1f3a9a12a0a9f646d63/dotfiles/.bash_functions#L64) comes in handy. I just open a terminal and call 'sl drake'.

wlandau:
It may be faster to [lint individual files with 'lt file.R'](https://github.com/wlandau/dotfiles/blob/94825dfdbc4243c0ff2bd1f3a9a12a0a9f646d63/dotfiles/.bash_functions#L35)

R/utils-deprecate.R
```diff
@@ -1390,3 +1390,84 @@ prune_drake_graph <- function(
...
 }
+
+#' @title Deprecated. Show the analysis wildcard
+#'   used in [plan_summaries()].
+#' @description Deprecated on 2019-01-12.
+#' @details Used to generate workflow plan data frames.
+#' @export
+#' @seealso [plan_summaries()]
+#' @return The analysis wildcard used in [plan_summaries()].
+analysis_wildcard <- function() { # nocov start
```
bpbond: @wlandau This `nocov` OK here and below? Otherwise we have coverage regression, or should I have a test for this function as well as `analysis_wildcard_`?

wlandau: It would be great to have a little test that calls it. I generally try to reserve `#nocov` for parts of the code that are too difficult or impractical to get at with automated unit tests. Examples:

- [`drake_debug()`](https://github.com/ropensci/drake/blob/a36c082b1c93017baf395efab24eb87fce235320/R/api-debug.R#L45), which pauses the code with `browser()`. Tested in `test-always-skipped.R` in interactive mode.
- Stuff predicated on modifying the global environment, like [here](https://github.com/ropensci/drake/blob/a36c082b1c93017baf395efab24eb87fce235320/R/backend-clustermq.R#L22). Unit tests should generally avoid accessing `globalenv()`. [`drake`'s larger testing suite](https://github.com/ropensci/drake/tree/master/tests/scenarios) has other ways to get at this.
- Lines whose execution is not actually guaranteed for workflows small enough to unit-test, like [here](https://github.com/ropensci/drake/blob/a36c082b1c93017baf395efab24eb87fce235320/R/backend-future.R#L24).
- In some cases, safety/efficiency checks that are no longer relevant but could become relevant again if other internals change (example [here](https://github.com/ropensci/drake/blob/a36c082b1c93017baf395efab24eb87fce235320/R/backend-clustermq.R#L63-L67)).

NAMESPACE
```diff
@@ -9,6 +9,7 @@ export(build_drake_graph)
 export(build_times)
 export(built)
 export(cache_namespaces)
+export(cache_namespaces_)
```
wlandau: Hmm... I wonder how this got exported. You rightly omitted an `@export` tag from the `roxygen2` markup.

R/api-cache.R
```diff
@@ -27,7 +28,7 @@ assert_cache <- function(cache) {
 #' cache_path(cache = mem)
 #' })
 #' }
-cache_path <- function(cache = NULL) {
+cache_path_ <- function(cache = NULL) {
```
wlandau: The functions with trailing underscores do not actually need `roxygen2` documentation since they are not part of the API.

R/utils-deprecate.R
```diff
@@ -1390,3 +1390,84 @@ prune_drake_graph <- function(
...
+  .Deprecated(
+    "analysis_wildcard",
+    package = "drake",
+    msg = "analysis_wildcard() is deprecated."
+  )
+  analysis_wildcard_()
```
wlandau: This is exactly what I had in mind for the functions you are working on. `analysis_wildcard()` is a deprecated API function whose only role is to throw a warning and call `analysis_wildcard_()`. `analysis_wildcard_()` should not be exported and does not need any `roxygen2` documentation, and it is used internally everywhere `analysis_wildcard()` used to be.

tests/testthat/test-dependencies.R
```diff
@@ -163,7 +163,7 @@ test_with_dir("tracked() works", {
 test_with_dir("missing input files", {
```
wlandau: I generally prefer these sorts of tests to test API functions rather than internal ones when possible. That tends to make the test suite less brittle. `drake_config()` should generate the expected warnings.

wlandau:
I think you are on the right track here, please see my inline comments. Also, would you maintain a list at the top of this thread with all the functions you plan to deprecate in this PR? So far I noticed these:

- `analysiss_wildcard()`
- `cache_namespaces()`
- `cache_path()`
- `check_plan()`

NAMESPACE
bpbond: 🤷‍♂️

R/utils-deprecate.R
bpbond: Makes sense.

tests/testthat/test-dependencies.R
wlandau: You know what? I just remembered that `check_plan()` is not actually used anywhere internally. (Other internals *did* use it in the distant past.) I think we can just get rid of `check_plan_()` and deprecate `check_plan()` the usual way. Sorry to have put you though extra trouble.

R/exec-imports.R
```diff
@@ -6,16 +6,9 @@ process_imports <- function(config) {
...
-#' @export
-#' @keywords internal
-#' @param import character, name of an import to process
-#' @param config [drake_config()] object
-#' @examples
-#' # Not a user-side function.
-process_import <- function(import, config) {
-  meta <- drake_meta(target = import, config = config)
+# Internal function only used inside process_imports(); not user-side.
+process_import_ <- function(import, config) {
```
wlandau: I forgot to tell you this, but `process_import()` actually needs to stay exported (with `@keywords internal` to hide the help file from the main index). The reason is that otherwise `parallel::parLapply()` cannot find it in R-3.3.0. There are other parallel computing workers like this.

R/utils-deprecate.R
```diff
@@ -1390,3 +1390,279 @@ prune_drake_graph <- function(
...
+
+#' @title Deprecated. List all the `storr` cache namespaces used by drake.
+#' @description Deprecated on 2019-01-12.
+#' @return A character vector of `storr` namespaces used for drake.
+#' @details Ordinary users do not need to worry about this function.
+#' It is just another window into `drake`'s internals.
+#' @param default name of the default `storr` namespace
+#' @export
+#' @seealso [make()]
+cache_namespaces <- function(
```
wlandau: From searching the code, I also realized that `cache_namespaces()` can be deprecated without a trailing underscore function. No internals rely on it.

tests/testthat/test-edge-cases.R
```diff
@@ -185,19 +185,6 @@ test_with_dir("stringsAsFactors can be TRUE", {
   expect_equal(readd(a), "helloworld")
 })
 
-# Target/import conflicts are unpredictable. A warning should
-# be enough.
-test_with_dir("target conflicts with current import or another target", {
```
wlandau: Even though `check_plan()` is deprecated, the underlying idea of this test is super important. It should be enough to replace `check_plan()` with `drake_config()`.

tests/testthat/test-edge-cases.R
```diff
@@ -185,19 +185,6 @@ test_with_dir("stringsAsFactors can be TRUE", {
...
-  skip_on_cran() # CRAN gets whitelist tests only (check time limits).
-  config <- dbug()
-  config$plan <- rbind(config$plan, data.frame(target = "f",
-    command = "1+1"))
-  expect_message(check_plan(plan = config$plan,
```
wlandau: Something I forgot to do: `expect_message(..., regexp = "Unloading targets from environment")`. would you add that `regexp` argument?

tests/testthat/test-other-features.R
```diff
@@ -188,26 +166,22 @@ test_with_dir(".onLoad() warns correctly and .onAttach() works", {
   expect_warning(drake:::.onLoad())
   unlink(f, force = TRUE)
   set.seed(0)
-  expect_true(is.character(drake_tip()))
+  expect_true(is.character(suppressWarnings(drake_tip())))
```
wlandau: Where possible, I would prefer to sequester all uses of deprecated API functions into `test-deprecate.R` This allows us to remove those functions a year from now without too much of a headache refactoring the test suite.

tests/testthat/test-other-features.R
```diff
@@ -15,7 +15,7 @@ test_with_dir("debug_command()", {
 test_with_dir("build_target() does not need to access cache", {
   skip_on_cran() # CRAN gets whitelist tests only (check time limits).
   config <- drake_config(drake_plan(x = 1), lock_envir = FALSE)
-  meta <- drake_meta(target = "x", config = config)
+  meta <- suppressWarnings(drake_meta(target = "x", config = config))
```
wlandau: Could we use `drake_meta_()` here? I know I expressed a preference to rely on API functions in tests, but since `drake_meta()` is deprecated, I would prefer all mentions of it to be in `test-deprecate.R`.

tests/testthat/test-edge-cases.R
bpbond:
```
Expected match: "Unloading targets from environment"
Actual message: "Duplicated targets with different commands:\n  repeated\n  repeated\n  repeated\n  repeated\n  repeated\n  repeated"
```
I have changed regexp to the latter?

tests/testthat/test-other-features.R
bpbond: What about the calls to `in_progress()` in `test-cache.R`? E.g.
```R
  expect_equal(suppressWarnings(in_progress(search = FALSE)), character(0))
```

tests/testthat/test-edge-cases.R
wlandau: Hmm, that's odd. With d3f1e58924a815cf5a62deb284577b32361922b8:

``` r
library(drake)
config <- drake:::dbug()
config$plan <- rbind(
  config$plan,
  data.frame(
    target = "f",
    command = "1+1"
  )
)
config2 <- drake_config(
  config$plan,
  envir = config$envir
)
#> Unloading targets from environment:
#>   f
```

<sup>Created on 2019-01-13 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

tests/testthat/test-other-features.R
wlandau: I think we could write `expect_false(any("in progress" %in% progress(search = FALSE)))`.

tests/testthat/test-edge-cases.R
bpbond: And with 3c0cbf3 in `hide-internals`:
``` r
library(drake)
config <- drake:::dbug()
config$plan <- rbind(config$plan, data.frame(target = "f", command = "1+1"))
config$plan$target <- "repeated"
drake_config(plan = config$plan)
#> Error: Duplicated targets with different commands:
#>   repeated
#>   repeated
#>   repeated
#>   repeated
#>   repeated
#>   repeated
```
I'll look into this.

bpbond: Oh wait...your original request above referred to the `expect_message` in line 194, not the `expect_error` in 197. Ahh. 👍

wlandau: Glad we got that sorted. Both tests are clutch.

wlandau:
Thanks, @bpbond! Like I said, your work to deprecate these functions really helps move development along. Downsizing the code base is a major goal.

Issue #702, opened 2019-02-01
Profile the DSL post-#696


wlandau:
Speed should have increased, but I should still verify that it did not decrease.

wlandau:
Just had a look. Nothing too egregious.

Issue #700, opened 2019-02-01
Treat commands as language objects


wlandau:
`drake` currently uses character strings to represent code.

```r
drake_plan(x = simulate_data())
#> # A tibble: 1 x 2
#>   target command        
#>   <chr>  <chr>          
#> 1 x      simulate_data()
```

Now that we have a DSL (#233) we can start to move away from text. Specific considerations:

1. Printing. https://github.com/r-lib/pillar/issues/34 has been quiet, so we may need a special S3 class and a `print()` method for `drake` plans. The trouble is that data frame operations often drop attributes.
2. Wildcard functions. In functions like `evaluate_plan()`, we should deparse the `command` column, do the text replacement, `safe_parse()` to get language objects, and then restore the S3 class. 
3. Back compatibility. Because this is such a large change, and because (1) may take time, I think I would prefer to delay this issue until after version 7.0.0 is released. That means `drake` still needs to work if commands are character strings and targets should stay valid.
- Command standardization. I think this will clean up nicely. We just need to remember to use `all.vars(functions = TRUE)` instead of `grepl()` to set `look_for_ignore` in `standardize_command()`.

wlandau:
Printing seems to work out okay now. Note the `<expr>` type summary below. Hopefully the deparsing is not too slow. I have not yet decided whether to discard the unnecessary deparsing or how best to do it.

```r
library(drake)
drake_plan(
  raw_data = readxl::read_excel(file_in("raw_data.xlsx")),
  data = raw_data %>%
    mutate(Species = forcats::fct_inorder(Species)),
  hist = create_plot(data),
  fit = lm(Sepal.Width ~ Petal.Width + Species, data),
  report = rmarkdown::render(
    knitr_in("report.Rmd"),
    output_file = file_out("report.html"),
    quiet = TRUE
  )
)
#> # A tibble: 5 x 2
#>   target   command                                                         
#>   <chr>    <expr>                                                          
#> 1 raw_data readxl::read_excel(file_in("raw_data.xlsx"))                   …
#> 2 data     raw_data %>% mutate(Species = forcats::fct_inorder(Species))   …
#> 3 hist     create_plot(data)                                              …
#> 4 fit      lm(Sepal.Width ~ Petal.Width + Species, data)                  …
#> 5 report   rmarkdown::render(knitr_in("report.Rmd"), output_file = file_ou…
```

Issue #699, opened 2019-02-01
Deprecate the `list` argument of drake_plan()


wlandau:

wlandau:
Fixed in 029be158cf011a415dcb9067a940e93e08f62199.

Issue #698, opened 2019-02-01
Deprecate target() as a standalone function


wlandau:
I no longer believe `target()` should be a user-side function. Especially with the DSL, it should only be called inside `drake_plan()`.

PR #703, opened 2019-02-01
Tidy evaluation in the DSL

### LINKED ISSUE #696: NSE applied within targets before transforms applied: file_in and file_out woes
My (bioinformatics) workflow involves application of command line utilities over lots of files. In order to do this in `drake`, my idea is to encapsulate each command line call in an R function.  Using standard unix utilities:

```r
my_copy <- function(from, to) {
system2("cp", args = c(from, to))
}

in_files <- paste0("in", 1:10, ".txt")
file.create(in_files)
out_files <- paste0("out", 1:10, ".txt")

plan <- drake_plan(
  copies = target(my_copy(file_in(the_in), file_out(the_out)),
                        transform = map(the_in = !!in_files, the_out = !!out_files))
# do more steps here...
)

make(plan)
```

This works fine.

However, some command line utilities I want to use have more than one input or output, which are not all explicitly listed on the command line (typically, a base name is given, and the command generates the full names from these). Exactly how many and which outputs will be created may depend on other options set at the command line.  I'll use [`split`](http://manpages.ubuntu.com/manpages/precise/en/man1/split.1.html) as an easily accessible example. The reasonable thing to do, I thought, was to use NSE to have the R function calculate what the input and output files would be, and then call `file_in` and `file_out` appropriately.  The function starts to look like a mess, but it works as intended.

```r
my_split <- function(from, stem, n) {
  suffixes <- with(expand.grid(y = letters, x = letters),
                   paste0(x,y))[1:n]
  out.files <- paste0(stem, suffixes)
  quo({
    file_in(!!from)
    file_out(!!out.files)
    system2("split", c(paste0("-n r/", !!n),
                       !!from,
                       !!stem))
  })
}

split_in = "longfile.txt"
writeLines(as.character(1:1000), split_in)

plan2 <- drake_plan(
  splits = target(!!my_split(split_in, "shortfile", 10))
)

make(plan2)
```

That is, it works as intended until I want to use a transform...

```r
manysplits <- paste0("longfile", 1:10, ".txt")
map(manysplits, writeLines, text = as.character(1:1000))

plan3 <- drake_plan(
  splits = target(!!my_split(f, f, 10),
                  transform = map(f = !!manysplits))
)
## Error in paste0(stem, suffixes) : object 'f' not found
```
Apparently `drake_plan` is applying NSE to the target (thus finding f) before it applies the transform, which would give `f` its value.  I'm sure sometimes this would be the desired effect, but in my case it's a big stumbling block.

wlandau:
# Summary

This PR fixes #696 so that tidy evaluation happens after transformation. The process was more complicated and involved than I expected, and it led to some new fixes.

- Wait until the very end of `drake_plan()` to turn commands into text. I consider it prework for #700.
- Apply transformations to triggers and other custom columns, not just the commands.
- Treat tidy evaluation in the other columns just like tidy evaluation in the commands.

And it raised some new issues that could delay the CRAN release of 7.0.0 for another month: #698, #699, #700, and #702.

- Ref: #696

# Checklist


wlandau:
cc @brendanf

Issue #724, opened 2019-02-08
`!!!` doesn't work in drake plan sometimes


tmastny:
This works 

``` r
library(drake)

get_data <- function(param) {
  mtcars
}

drake_plan(
  data = target(
    get_data(param),
    transform = map(param = c(1, 2, 3))
  ),
  results = target(
    bind_rows(!!!data), 
    transform = combine(data)
  )
)
#> # A tibble: 4 x 2
#>   target  command                          
#>   <chr>   <expr>                           
#> 1 data_1  get_data(1)                      
#> 2 data_2  get_data(2)                      
#> 3 data_3  get_data(3)                      
#> 4 results bind_rows(data_1, data_2, data_3)
```

<sup>Created on 2019-02-08 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

```
> sessionInfo()
R version 3.5.1 (2018-07-02)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS  10.14.2

Matrix products: default
BLAS: /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] drake_6.2.1.9003

loaded via a namespace (and not attached):
 [1] fansi_0.4.0         assertthat_0.2.0    utf8_1.1.4          packrat_0.5.0      
 [5] digest_0.6.18       crayon_1.3.4        R6_2.3.0            backports_1.1.3    
 [9] storr_1.2.1         magrittr_1.5.0.9000 pillar_1.3.1        cli_1.0.1          
[13] rlang_0.3.1         rstudioapi_0.9.0    tools_3.5.1         igraph_1.2.2       
[17] compiler_3.5.1      pkgconfig_2.0.2     base64url_1.4       tibble_2.0.1   
```

But on another machine:

```r
library(drake)

get_data <- function(param) {
  mtcars
}

drake_plan(
  data = target(
    get_data(param),
    transform = map(param = c(1, 2, 3))
  ),
  results = target(
    bind_rows(!!!data),
    transform = combine(data)
[...]

PR #729, opened 2019-02-10
Custom grouping functions in the DSL

wlandau:
# Summary

In the current DSL, `combine(stuff)` always gathers targets into lists, which is too limiting. This PR allows users to supply custom grouping functions. 

``` r
library(drake)
drake_plan(
  data = target(
    get_data(param),
    transform = map(param = c(1, 2))
  ),
  results = target(
    .data %>%
      data,
    transform = combine(data = select)
  )
)
#> # A tibble: 3 x 2
#>   target  command                         
#>   <chr>   <expr>                          
#> 1 data_1  get_data(1)                     
#> 2 data_2  get_data(2)                     
#> 3 results .data %>% select(data_1, data_2)
```

<sup>Created on 2019-02-10 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

``` r
library(drake)
drake_plan(
  data = target(
    get_data(param),
    transform = map(param = c(1, 2))
  ),
  results = target(
    .data %>%
      data,
    transform = combine(data = min(0, na.rm = FALSE))
  )
)
#> # A tibble: 3 x 2
#>   target  command                                        
#>   <chr>   <expr>                                         
#> 1 data_1  get_data(1)                                    
#> 2 data_2  get_data(2)                                    
#> 3 results .data %>% min(data_1, data_2, 0, na.rm = FALSE)
```

<sup>Created on 2019-02-10 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

The following is equivalent to the current DSL.

``` r
library(drake)
drake_plan(
  data = target(
    get_data(param),
    transform = map(param = c(1, 2))
  ),
  results = target(
    .data %>%
      data,
    transform = combine(data = list())
  )
)
#> # A tibble: 3 x 2
#>   target  command                       
#>   <chr>   <expr>                        
#> 1 data_1  get_data(1)                   
#> 2 data_2  get_data(2)                   
#> 3 results .data %>% list(data_1, data_2)
```

<sup>Created on 2019-02-10 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

One new restriction: you must always supply a grouping call to every grouping variable.

``` r
library(drake)
drake_plan(
  data = target(
    get_data(param),
    transform = map(param = c(1, 2))
  ),
  results = target(
    .data %>%
      data,
    transform = combine(data)
  )
)
#> Error: please supply a grouping function to each grouping variable in combine(), e.g. combine(data = list()) instead of combine(data).
```

<sup>Created on 2019-02-10 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

This API is not as powerful or flexible as tidy evaluation's unquote-splice operator (`!!!`) but it is still very powerful and it fits `drake`'s DSL very nicely.

cc @tmastny 

- Ref: #724

# Checklist

### DIFF
--- a/R/api-dsl.R
+++ b/R/api-dsl.R
@@ -137,7 +137,7 @@ dsl_transform.cross <- dsl_transform.map <- map_to_grid
 
 dsl_transform.combine <- function(transform, target, row, plan) {
   old_cols <- old_cols(plan)
-  cols_keep <- union(dsl_by(transform), dsl_combine(transform))
+  cols_keep <- union(dsl_by(transform), names(dsl_combine(transform)))
   rows_keep <- complete_cases(plan[, cols_keep, drop = FALSE])
   if (!length(rows_keep) || !any(rows_keep)) {
     row[["transform"]][[1]] <- NA
@@ -158,24 +158,33 @@ dsl_transform.combine <- function(transform, target, row, plan) {
 }
 
 combine_step <- function(plan, row, transform, old_cols) {
-  aggregates <- lapply(
-    X = plan[, dsl_combine(transform), drop = FALSE],
-    FUN = function(x) {
-      unname(lapply(as.character(na_omit(unique(x))), as.symbol))
-    }
-  )
+  env <- env_combine(plan, transform)
   out <- data.frame(command = NA, stringsAsFactors = FALSE)
   for (col in setdiff(old_cols, c("target", "transform"))) {
-    out[[col]] <- list(substitute_list(row[[col]][[1]], aggregates))
+    out[[col]] <- list(
+      eval(call("substitute", row[[col]][[1]], env), envir = baseenv())
+    )
   }
   out
 }
 
-substitute_list <- function(expr, env) {
-  env <- lapply(env, function(lst) {
-    as.call(c(quote(list), lst))
-  })
-  eval(call("substitute", expr, env), envir = baseenv())
+env_combine <- function(plan, transform) {
+  out <- lapply(
+    names(dsl_combine(transform)),
+    env_combine_entry,
+    plan = plan,
+    transform = transform
+  )
+  names(out) <- names(dsl_combine(transform))
+  out
+}
+
+env_combine_entry <- function(name, transform, plan) {
+  grouping_call <- dsl_combine(transform)[[name]]
+  levels <- unname(
+    lapply(as.character(na_omit(unique(plan[[name]]))), as.symbol)
+  )
+  as.call(c(grouping_call[[1]], levels, as.list(grouping_call[-1])))
 }
 
 lang <- function(...) UseMethod("lang")
@@ -257,7 +266,7 @@ dsl_deps.map <- dsl_deps.cross <- function(transform) {
 
 dsl_deps.combine <- function(transform) {
   attr(transform, "deps") %|||% c(
-    as.character(unnamed(transform[[1]][-1])),
+    names(dsl_combine(transform)),
     dsl_by(transform)
   )
 }
@@ -272,8 +281,28 @@ dsl_by.combine <- function(transform) {
 dsl_combine <- function(...) UseMethod("dsl_combine")
 
 dsl_combine.combine <- function(transform) {
-  attr(transform, "combine") %|||%
-    as.character(unnamed(lang(transform))[-1])
+  if (!is.null(attr(transform, "combine"))) {
+    return(attr(transform, "combine"))
+  }
+  expr <- lang(transform)
+  if (length(as.character(unnamed(expr)[-1]))) {
+    stop(
+      "please supply a grouping function to each grouping variable in ",
+      "combine(), e.g. combine(data = list()) instead of combine(data).",
+      call. = FALSE
+    )
+  }
+  named <- named(as.list(expr))
+  names <- names(named)
+  named <- named[setdiff(names(named), c(".by", dsl_all_special))]
+  named <- lapply(named, function(x) {
+    if (is.symbol(x)) {
+      x <- as.call(c(x))
+    }
+    x
+  })
+  names(named) <- names
+  named
 }
 
 new_groupings <- function(...) UseMethod("new_groupings")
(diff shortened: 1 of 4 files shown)


wlandau:
The AppVeyor builds are failing because `igraph` did not compile: https://ci.appveyor.com/project/wlandau/drake/builds/22259081#L1602. We will have failing checks until the CRAN builds are available for Windows.

tmastny:
The API seems a little idiosyncratic. In most R contexts, `combine(data = select)`this implies that `data` is being assigned `select`, or `select` is now the parameter for the data argument.

It might be more consistent if we thought about `combine` like a function:
```
combine <- function(data, combine_function) ...
```
so the api would look like 
```
combine(data, select)
combine(data, list)
combine(data, c)
```

wlandau:
I see what you mean. The best thing would have been native unquote-splicing like `!!!`, but I do not think that is available to us.

The trouble is that `combine()` needs a way to create multiple variable-function pairs, and `combine(data, select)` only accounts for a single pair. Multiple `combine()` calls in a single transform would be too complicated internally, and `combine(select = data1, select = data2)` creates names conflicts and precludes the example with `min(0, na.rm = FALSE)`.

If documented properly, I think we can work with a bit of strangeness. Here, we are supplying *rules* to combine targets, not the targets themselves. In that sense, this PR implements a sort-of inverse of the `rules` argument to `evaluate_plan()`.

tmastny:
Along those same lines, what if we needed to pass additional parameters to our combine function?

Our combine function might have additional parameters we need to supply, as well as values we need to pass to `...`.

For example, in this case `gather` requires key and value parameters, and then some number of column names passed to `...`:
``` r
library(tidyverse)
iris %>%
  gather(key = "type", value = "dimension", Sepal.Length, Sepal.Width) %>%
  as_tibble()
#> # A tibble: 300 x 5
#>    Petal.Length Petal.Width Species type         dimension
#>           <dbl>       <dbl> <fct>   <chr>            <dbl>
#>  1          1.4         0.2 setosa  Sepal.Length       5.1
#>  2          1.4         0.2 setosa  Sepal.Length       4.9
#>  3          1.3         0.2 setosa  Sepal.Length       4.7
#>  4          1.5         0.2 setosa  Sepal.Length       4.6
#>  5          1.4         0.2 setosa  Sepal.Length       5  
#>  6          1.7         0.4 setosa  Sepal.Length       5.4
#>  7          1.4         0.3 setosa  Sepal.Length       4.6
#>  8          1.5         0.2 setosa  Sepal.Length       5  
#>  9          1.4         0.2 setosa  Sepal.Length       4.4
#> 10          1.5         0.1 setosa  Sepal.Length       4.9
#> # … with 290 more rows
```

<sup>Created on 2019-02-10 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

wlandau:
That's actually one reason I like the existing implementation of #729. You can absolutely do that here.

``` r
library(drake)
drake_plan(
  data = target(
    get_data(param),
    transform = map(param = c(1, 2))
  ),
  results = target(
    .data %>%
      data,
    transform = combine(data = min(0, na.rm = FALSE))
  )
)
#> # A tibble: 3 x 2
#>   target  command                                        
#>   <chr>   <expr>                                         
#> 1 data_1  get_data(1)                                    
#> 2 data_2  get_data(2)                                    
#> 3 results .data %>% min(data_1, data_2, 0, na.rm = FALSE)
```

<sup>Created on 2019-02-10 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

tmastny:
Ah, of course! I didn't read it closely enough

wlandau:
No worries. Like you said, it's a strange approach.

tmastny:
Oh wait, but in that example it splices the value at the start of `min`. In gather's case, it needs to be add the end. How does that work?

wlandau:
Really? It works for me...

``` r
library(tidyverse)
iris %>%
  gather(Sepal.Length, Sepal.Width, key = "type", value = "dimension") %>%
  as_tibble()
#> # A tibble: 300 x 5
#>    Petal.Length Petal.Width Species type         dimension
#>           <dbl>       <dbl> <fct>   <chr>            <dbl>
#>  1          1.4         0.2 setosa  Sepal.Length       5.1
#>  2          1.4         0.2 setosa  Sepal.Length       4.9
#>  3          1.3         0.2 setosa  Sepal.Length       4.7
#>  4          1.5         0.2 setosa  Sepal.Length       4.6
#>  5          1.4         0.2 setosa  Sepal.Length       5  
#>  6          1.7         0.4 setosa  Sepal.Length       5.4
#>  7          1.4         0.3 setosa  Sepal.Length       4.6
#>  8          1.5         0.2 setosa  Sepal.Length       5  
#>  9          1.4         0.2 setosa  Sepal.Length       4.4
#> 10          1.5         0.1 setosa  Sepal.Length       4.9
#> # … with 290 more rows
```

<sup>Created on 2019-02-10 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

wlandau:
Taking `lapply()` and `purrr::map()` as precedent, I think users will expect spliced arguments to go first.

tmastny:
Your case works because you directly supplied the parameter names. If you did it implicitly, the order would matter:
```
gather(data, key = "key", value = "value", ..., na.rm = FALSE,
  convert = FALSE, factor_key = FALSE)
```
I think that's fine, just something to be aware of. 


Additionally, it looks like in purrr the splicing goes last as well: https://purrr.tidyverse.org/reference/map.html

wlandau:
Yeah, you are right about `lapply()` and `map()`. I am not actually sure there is a good default here. `min()`, `max()`, and `bind_rows()` do splicing first, while `select()`, `gather()`, `mutate()`, etc. do splicing last. But in this case, I like splicing first because those grouping variables are most likely going to be the center of attention. Plus, it may be a good thing to nudge users to name their arguments.

wlandau:
Update: #730 undoes #729 with actual splicing (which is what we wanted anyway).

PR #798, opened 2019-03-25
rotemplate was transferred

maelle:


wlandau:
Congrats! I will merge when the builds complete.

PR #826, opened 2019-04-08
make_spinner requires cli >=1.1.0

billdenney:
I just tried to install `drake` from CRAN which required compilation for windows.  It gave an error that I tracked down to needing to update `cli`.  I added the version requirement to the Suggests for the version of `cli` required to install (because of `make_spinner()`).


wlandau:
Thanks!

Issue #945, opened 2019-07-16
Recover targets from the distant past to avoid rerunning commands


wlandau:
## Prework

- [x] Read and abide by `drake`'s [code of conduct](https://github.com/ropensci/drake/blob/master/CODE_OF_CONDUCT.md).
- [x] Search for duplicates among the [existing issues](https://github.com/ropensci/drake/issues), both open and closed.

## Background

Let's start with a simple plan.

```r
library(drake)
library(tidyverse)

plan <- drake_plan(
  x = iris %>%
    group_by(Species) %>%
    summarize_all(mean)
)

make(plan)
# target x
```

As we work on the project, we make changes. For example:

```r
library(drake)
library(tidyverse)

plan <- drake_plan(
  x = iris %>%
    group_by(Species) %>%
    summarize_all(sum) # changed from `mean` to `sum`
)

make(plan)
# target x
```

## Proposal

Suppose we go back to our original plan, e.g. using `mean` instead of `sum`. The current `drake` reruns command and builds the target. But unless we ran `clean(garbage_collection = TRUE)`, that old data is still in the cache. What if we restored the old data instead of rerunning the command?

```r
library(drake)
library(tidyverse)

plan <- drake_plan(
  x = iris %>%
    group_by(Species) %>%
    summarize_all(mean)
)

make(plan)
# target x
# restore x
```

wlandau:
We need to recover the metadata hash from the old build. Could be tricky.

wlandau:
## Good news

I think this is possible. And for the implementation I have in mind, the target's name will not matter. In fact, `rename_targets()` should no longer be necessary (#936). Simply changing the name of the target should tell `drake` to salvage old data, not compute new data.

## Bad news

The implementation will inevitably create more tiny files in the cache. If you are worried about maxing out the file quota on your machine, this could be a problem. Maybe we should have a new logical `recoverable` argument in `make()` and `drake_config()`.

wlandau:
And, unavoidably, renaming a target will always invalidate downstream targets because of the way `drake` computes dependency hashes.

wlandau:
Another obstacle: seeds. A target's seed is determined by the global seed and the target name. This is also a problem for the existing `rename_targets()`. Can we

1. Find a new method to come up with a target's seed?
2. Roll out the new method without breaking existing workflows?

wlandau:
[...]

PR #952, opened 2019-07-18
Automated data recovery

wlandau:
# Summary

This PR is for discussion only. I am not sure it would be best to follow through just yet. It may or may not end up as an experimental feature in the next release. Either way, I would really like to learn what users think.

When you call `clean()` without garbage collection (default) the actual target data is still in the cache. That means we can implement automated data recovery. We can skip a target's command if the dependencies etc. are shared by an old target whose value we still have.

``` r
library(drake)
library(tidyverse)

plan <- drake_plan(
  x = iris %>%
    group_by(Species) %>%
    summarize_all(mean)
)

make(plan)
#> target x

# Invalidate the target.
clean()

# The data is still there.
make(plan, recover = TRUE)
#> recover x

# Change x.
plan <- drake_plan(
  x = iris %>%
    group_by(Species) %>%
    summarize_all(sum)
)

make(plan)
#> target x

# Change it back.
plan <- drake_plan(
  x = iris %>%
    group_by(Species) %>%
    summarize_all(mean)
)

make(plan, recover = TRUE)
#> recover x

# If we set a custom seed,
# we can rename the target
# without rebuilding it.
plan <- drake_plan(
  x = target(
    iris %>%
      group_by(Species) %>%
      summarize_all(sum),
    seed = 123
  )
)

# `recover` is TRUE, but the dependencies do not match
# those of a previous build in the cache.
make(plan, recover = TRUE)
#> target x

plan <- drake_plan(
  y = target(
    iris %>%
      group_by(Species) %>%
      summarize_all(sum),
    seed = 123
  )
)

make(plan, recover = TRUE)
#> recover y
```

<sup>Created on 2019-07-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

But this approach has problems.

1. Reproducibility. We do not know how far back these old targets go. We might be pulling data from old R versions and package environments that no longer exist.
2. Performance. For a [large number of empty targets](https://github.com/wlandau/drake-examples/tree/master/overhead), the extra bookkeeping incurs a 7.5% runtime penalty. Plus, the implementation requires lots of tiny files that encumber the cache.

- Ref: #945

# Checklist


wlandau:
Another issue: the implementation requires direct access to `config$cache$driver$set_hash()`, which is not part of the unified `storr` API. We have `set_hash()` in the RDS, environment, and DBI drivers (which I think are the ones we care about the most) but not necessarily in all drivers. `drake` does check if we have a `set_hash()` method before we try to recover old targets. But is it still okay to use a driver method directly? cc @richfitz.

richfitz:
`set_hash` is part of the `storr` spec, so all drivers should have it with no need to check: https://github.com/richfitz/storr/blob/5c5475224b7311128e2fcee456280fa7b97499b8/inst/spec/test-driver.R#L31-L70

The difference is in what these functions will accept, which you can get by querying the drivers traits.

wlandau:
Awesome, thanks for clarifying. What about `hash_algorithm`? It is a trait of the more commonly used drivers, but I do not see it in the spec.

richfitz:
https://github.com/richfitz/storr/blob/5c5475224b7311128e2fcee456280fa7b97499b8/inst/spec/test-storr.R#L185-L229 - looks like I never made it a required trait.  So you can test the `$hash_algorithm` logical trait to see if it's supported

wlandau:
Hmm... I suppose I will have to enforce it in `drake`. Glad I know now.

wlandau:
Back to the main thread: after some more commits, I think the main concerns are mostly alleviated.

1. Reproducibility. The latest commit takes some precautions. `make(recover = FALSE, recoverable = TRUE)` is the default. The data is automatically indexed for potential recovery, but you have to opt into the actually take advantage of the recovery itself. Once you do, you can see when a target's data was actually stored using `diagnose(my_target)$date`.
2. Speed. I refactored the implementation to avoid saving extra stuff to `.drake/data/`, which shaved off 50% of the overhead. Now, it takes less time to make data recoverable than it does to track progress.
3. `storr` API: we are allowed to use `set_hash()`.

Let's make this an experimental feature in the next release.

PR #982, opened 2019-08-08
Avoid downloading data when rehashing URL

noamross:
# Summary

It seems that current behavior for rehashing  URL still downloads the whole contents of that URL to memory.  This change avoids doing so at all by telling `curl` to only get response headers, no body.

I'm not sure where or how in the test framework to set up a test for "Does not download the whole URL content body when testing for changes at the URL."

- Ref: #

# Checklist

- [ ] I have added [`testthat`](https://github.com/r-lib/testthat) unit tests to [`tests/testthat`](https://github.com/ropensci/drake/tree/master/tests/testthat) for any new functionality.
    -   I have made an in-line runtime test condition.


wlandau:
Thanks for catching this, @noamross. Would you mention it in `NEWS.md`?

For testing, I think we could [mock](https://github.com/krlmlr/mockr) `assert_useful_headers()` in `rehash_url()` or `assert_useful_headers()` to make sure we download just the headers.

wlandau:
Hmm... instead of `mockr`, what about `stopifnot(length(req$content) < 1L)` just after `req <- curl::curl_fetch_memory(url,  handle = curl::new_handle(nobody = TRUE))`?

wlandau:
A couple more comments:

1. Don't worry about the Travis failure. I am almost positive it is caused by https://github.com/r-lib/covr/issues/392. Feel free to switch back to CRAN `covr` in `.travis.yml`.
2. If you would like, please feel free to add yourself as a contributor in the DESCRIPTION. It's a small PR, but it uncovered an important performance issue.

wlandau:
In fact, for (1), I will rerun Travis since the issue with `covr` should now be fixed.

wlandau:
Looks great, Noam. Thanks again!

PR #1013, opened 2019-09-14
make `seed_trigger` test for equality instead of identity

CreRecombinase:
# Summary

After trying to figure out why a large number of my targets were out date I eventually discovered that for all of the "outdated" targets the seed stored by the cache was of type `double`  while the seed returned from `drake_meta_` was type `integer` (they were otherwise equal). I'm guessing this is due to me updating drake at some point.  Because `seed_trigger` currently uses `identical` to compare seeds, these targets were invalidated despite not having changed. 

I should probably have made an issue about the erroneous invalidation of targets, but I don't think there's much hope of me being able to create a reprex.

- Ref: #

# Checklist

### DIFF
--- a/R/handle_trigger.R
+++ b/R/handle_trigger.R
@@ -178,7 +178,7 @@ seed_trigger <- function(target, meta, config) {
     field = "seed",
     cache = config$cache
   )
-  seed != meta$seed
+  !identical(as.integer(seed),  as.integer(meta$seed))
 }
 
 condition_trigger <- function(target, meta, config) {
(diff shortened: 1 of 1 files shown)


wlandau:
Thanks for investigating. Discovering the root of the problem could not have been easy...

Could we do `!identical(as.integer(seed), as.integer(meta$seed))` instead of `seed!=meta$seed`? That way, a `seed` not of length 1 should not cause strange errors.

wlandau:
Thanks!

PR #1165, opened 2020-02-08
Improvements to logging and console output

wlandau:
# Summary

This PR implements several improvements to logging.

## progress bar

`verbose = 2` in `make()` now has a different meaning. The spinner for preprocessing is gone, and now we have a progress bar for targets. Simplifications to infrastructure make this possible. cc @kendonB. 

```r
plan <- drake_plan(
  x = target(
    Sys.sleep(.1),
    transform = map(y = !!seq_len(40))
  ),
  w = seq_len(40),
  z = target(
    Sys.sleep(.1),
    dynamic = map(y = w)
  )
)

make(plan)
targets [==============================>-----]  86%
```

## cli symbols instead of colors

`drake`'s default line-by-line target-level console messages have always had a color scheme that was awkward and not optimized for white terminals. Now, we are using `cli` symbols and only the most basic colors for the symbols. The text is always the default color now.

``` r
library(drake)
plan <- drake_plan(
  a = 1,
  b = cancel(),
  c = stop(),
  d = target(stop(), retries = 2),
  e = seq_len(2),
  f = target(e, dynamic = map(e))
)

make(plan, keep_going = TRUE)
#> ▶ target a
#> ▶ target b
#> ■ cancel b
#> ▶ target c
#> x fail c
#> ▶ target d
#> ! retry d
#> ! retry d
#> x fail d
#> ▶ target e
#> ▶ dynamic f
#> > subtarget f_0b3474bd
#> > subtarget f_b2a5c9b8
#> ■ aggregate f

plan <- drake_plan(a = 1)

make(plan)
#> ✓ All targets are already up to date.
```

<sup>Created on 2020-02-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

## console_log_file

The `console_log_file` argument of `make()` has an awkward name. This PR deprecates it and renames it `log_make`.

# Checklist

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -136,6 +136,7 @@ Suggests:
   lubridate,
   networkD3,
   prettycode,
+  progress (>= 1.2.2),
   qs (>= 0.20.2),
   Rcpp,
   rmarkdown,
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,7 +6,10 @@
 
 ## Enhancements
 
-* Assert dependencies of formats at the very beginning of `make()`, not in `drake_config()` (#1156). 
+* Assert dependencies of formats at the very beginning of `make()`, not in `drake_config()` (#1156).
+* In `make(verbose = 2)`, remove the spinner and use a progress bar to track how many targets are done so far.
+* Reduce logging of utility functions.
+* Improve the aesthetics of console messages using `cli` (optional package).
 
 # Version 7.10.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,7 @@
 * In `make(verbose = 2)`, remove the spinner and use a progress bar to track how many targets are done so far.
 * Reduce logging of utility functions.
 * Improve the aesthetics of console messages using `cli` (optional package).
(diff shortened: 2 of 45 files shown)


kendonB:
A progress bar is great! I think that verbose = 1 is now more verbose than verbose = 2 so I suggest they be switched. I also notice that subtargets don't tick the progress bar. Is that by design? I would have expected subtargets to tick the progress bar.

wlandau:
> A progress bar is great!

Glad you think so!

>  I think that verbose = 1 is now more verbose than verbose = 2 so I suggest they be switched. 

I will think about it, but some people might still supply the old `verbose = TRUE` and expect the old default. (And mapping `TRUE` to 2 instead of 1 may be counterintuitive.)

>  I also notice that subtargets don't tick the progress bar. Is that by design? I would have expected subtargets to tick the progress bar.

Sub-targets actually do tick the progress bar. You can try for yourself with one of the interactive tests.

https://github.com/ropensci/drake/blob/618b2f3ceb791b153ded30988c88cbc85350d0e0/tests/testthat/test-1-interactive.R#L400-L411

The progress bar almost reaches the end when it builds the `x_*` targets (static) and then backpedals when it reaches `z` (dynamic). It has to do this because it does not know how many sub-targets there will be in advance.

Issue #1268, opened 2020-06-04
cached_planned() and cached_unplanned() not working with non-standard cache locations


Plebejer:
## Prework

- [x] Read and abide by `drake`'s [code of conduct](https://github.com/ropensci/drake/blob/master/CODE_OF_CONDUCT.md).
- [x] Search for duplicates among the [existing issues](https://github.com/ropensci/drake/issues), both open and closed.
- [x] Advanced users: verify that the bug still persists in the current development version (i.e. `remotes::install_github("ropensci/drake")`) and mention the [SHA-1 hash](https://git-scm.com/book/en/v1/Getting-Started-Git-Basics#Git-Has-Integrity) of the [Git commit you install](https://github.com/ropensci/drake/commits/master). Hash: [53214ac9a34dbc7102a1dd87fbc79ad896de9131](https://github.com/ropensci/drake/commit/53214ac9a34dbc7102a1dd87fbc79ad896de9131).

## Description

If the drake cache is not in the standard location, the functions `cached_planned()` and `cached_unplanned()` do not work.

## Reproducible example

``` r
library(drake)

# create cache
cache_path <- file.path(tempdir(), ".drake")
my_cache <- new_cache(cache_path)

# create and make plan
plan <- drake_plan(iris = iris)
make(plan, cache = my_cache)
#> ▶ target iris

# change and remake plan
plan <- drake_plan(mtcars = mtcars)
make(plan, cache = my_cache)
#> ▶ target mtcars

# show error
cached_unplanned(plan, cache = my_cache)
Error: $ operator is invalid for atomic vectors
```

<sup>Created on 2020-06-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

## Expected result

This should work, independently from the cache location.

## Session info

<details>

<summary>Session info</summary>

``` r
devtools::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value                       
#>  version  R version 4.0.0 (2020-04-24)
#>  os       elementary OS 5.1.5 Hera    
#>  system   x86_64, linux-gnu           
#>  ui       X11                         
#>  language EN                          
#>  collate  de_DE.UTF-8                 
#>  ctype    de_DE.UTF-8                 
#>  tz       Europe/Berlin               
#>  date     2020-06-04                  
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version     date       lib source                            
[...]

PR #1269, opened 2020-06-04
cached_planned() and cached_unplanned() with cache in non-standard location

Plebejer:
# Summary

`cached_(un)planned()` did not work with caches in non-standard locations because the cache argument was not passed down to `subtargets()`. Thus, `subtargets()` defaulted to the default cache location. With this PR, the cache argument is passed to `subtargets()`.

- Ref: #1268 

# Checklist

### DIFF
--- a/tests/testthat/test-8-cache.R
+++ b/tests/testthat/test-8-cache.R
@@ -960,3 +960,24 @@ test_with_dir("drake_cancelled() (#1205)", {
   make(plan)
   expect_equal(drake_cancelled(), "y")
 })
+
+test_with_dir("cached_(un)planned() with custom cache location (#1268)", {
+  skip_on_cran()
+  path_cache <- file.path(tempdir(), ".drake")
+  test_cache <- new_cache(path_cache)
+  plan <- drake_plan(w = 1)
+  make(plan, cache = test_cache)
+  expect_equal(cached_planned(plan, cache = test_cache), "w")
+  expect_equal(cached_unplanned(plan, cache = test_cache), character(0))
+  plan <- drake_plan(
+    x = 2
+  )
+  expect_equal(cached_planned(plan, cache = test_cache), character(0))
+  expect_equal(cached_unplanned(plan, cache = test_cache), "w")
+  make(plan, cache = test_cache)
+  expect_equal(cached_planned(plan, cache = test_cache), "x")
+  expect_equal(cached_unplanned(plan, cache = test_cache), "w")
+  expect_equal(sort(cached(cache = test_cache)), sort(c("w", "x")))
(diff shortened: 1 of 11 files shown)


tests/testthat/test-8-cache.R
```diff
@@ -960,3 +960,24 @@ test_with_dir("drake_cancelled() (#1205)", {
...
+  plan <- drake_plan(
+    x = 2
+  )
+  expect_equal(cached_planned(plan, cache = test_cache), character(0))
+  expect_equal(cached_unplanned(plan, cache = test_cache), "w")
+  make(plan, cache = test_cache)
+  expect_equal(cached_planned(plan, cache = test_cache), "x")
+  expect_equal(cached_unplanned(plan, cache = test_cache), "w")
+  expect_equal(sort(cached(cache = test_cache)), sort(c("w", "x")))
+  clean(list = cached_unplanned(plan, cache = test_cache))
```
wlandau, line 981: `clean()` also needs a `cache` argument, assuming the intent is to clean `test_cache`. If that is the case, `cached()` should be "x" in the line below.

Plebejer: Thanks for pointing that out. I'll change that. 

Regarding the other failing tests: I'm not sure what to do about them since these are (should) not be affected by my patch. Am I wrong? 

BTW: Thanks for your patience with me in the last couple of weeks! This encouraged me to prepare my first PR to a project which is not my own (this one).

tests/testthat/test-8-cache.R
```diff
@@ -960,3 +960,22 @@ test_with_dir("drake_cancelled() (#1205)", {
   make(plan)
   expect_equal(drake_cancelled(), "y")
 })
+
+test_with_dir("cached_(un)planned() with custom cache location (#1268)", {
+  skip_on_cran()
+  path_cache <- tempfile()
```
wlandau, line 966: This line used to be `file.path(tempdir(), ".drake")`, which was causing problems. `tempdir()` is the parent directory of all the `tempfile()`s, and every test is in a `tempfile()` (via `test_with_dir()`). So every test after this one was seeing a `.drake` cache. Note: non-standard caches can have any folder name, not necessarily ".drake".

wlandau:
I think I figured out the problem with those tests (see https://github.com/ropensci/drake/pull/1269/files#r436101522). Since `test_with_dir()` creates a new directory at `tempfile()`, having a `drake` cache at `file.path(tempdir(), ".drake")` was problematic since `tempdir()` sits above `tempfile()` and so all the subsequent tests were seeing a preexisting `.drake` cache.

I will merge once the tests pass. Thanks again for the contribution!

